PR #104, opened 2023-07-17
sound stop function, board diagram updates

gpstar81:


source/NeutronaWand/Header.h
```diff
@@ -305,6 +305,13 @@ bool b_sound_firing_cross_the_streams = false;
 bool b_sound_idle = false;
 bool b_beeping = false;
 
+/*
```
DustinGrau, line 308: Nice, I should've looked to see if Arduino supported prototypes or "forward" declarations. Didn't realize it did.

source/NeutronaWand/NeutronaWand.ino
```diff
@@ -5285,6 +5236,59 @@ void wandSerialSend(int i_message) {
...
+    w_trig.trackGain(i_track_id, i_track_volume);
+    w_trig.trackPlayPoly(i_track_id, true);
+  }
+
+  if(b_track_loop == true) {
+    w_trig.trackLoop(i_track_id, 1);
+  }
+}
+
+void stopEffect(int i_track_id) {
```
DustinGrau, line 5263: I tried looking at passing an array of tracks to stop, but the code was getting too complex and Arduino doesn't seem to like creating an array on the fly due to strong typing. This is still better than just the direct WavTrigger calls throughout the code (and abstracts the audio player away from the code)

source/ProtonPack/Header.h
```diff
@@ -298,6 +298,13 @@ bool b_switch_mode_override = false; // Year mode override flag controlled by th
 bool b_pack_on = false;
 bool b_pack_shutting_down = false;
 
+/*
+ * Function prototypes for handling sound effects.
+*/
+void playEffect(int i_track_id, bool b_track_loop = false, int8_t i_track_volume = i_volume, bool b_fade_in = false, unsigned int i_fade_time = 0);
```
DustinGrau, line 304: I've just noticed that both headers do not include a prototype for the stopEffect function, but I assume that should be added?

DustinGrau (changes requested):
Only item I could find appears to be a missing function prototype for the new stopEffect in the pack/wand header files.

source/ProtonPack/Header.h
gpstar81: Oops good change, I will add it.

source/NeutronaWand/Configuration.h
```diff
@@ -79,7 +79,7 @@ const uint8_t VOLUME_EFFECTS_MULTIPLIER = 5;
  * Part #: BL28Z-3005SA04Y
  * Only compatible with the gpstar Neutrona Wand board, and not a Arduino Nano.
 */
-bool b_28segment_bargraph = false;
+bool b_28segment_bargraph = true;
```
DustinGrau, line 82: This just came up as a change--was this intentional as the new norm, or just part of other testing?

DustinGrau (approved):
Approved, but with the caveat/question as to whether changes to the config for the bargraph were intentional or not. This benefits me and I'm fine with new defaults, but wanted to be certain. I would suppose this is in response to preorders which may have overwhelmingly asked for the upgraded bargraph.

source/NeutronaWand/Configuration.h
gpstar81: ah damn, that should be false. I will force change it.

PR #109, opened 2023-07-23
Updated Color System

DustinGrau:
This is a fairly significant update to the way colors are handled for the LED's. The major change is a switch from defining as RGB values and using the HSV for hue (color), saturation, and brightness. The primary benefit being that there is no longer a need to calculate the individual values with respect to brightness, allowing that value to be passed directly to the new getHue() function, and using hsv2rgb to convert when necessary (as is typical with LED's which require a GRB color order). The other impact is the way the colors are defined which is via an ENUM type allowing for not only solid colors but overall "schemes". One example (not yet implemented) is a color chase which demonstrates how you could set a single value to indicate this scheme and every call to the getHue() will return the next in sequence.

As part of the change in how colors are handled, there are now 3 events which can be combined for actions:
1. Getting the brightness for a device, translating a percentage to the expected range for the FastLED library.
2. Getting the color scheme for a device, which is based on available system/firing modes and toggles.
3. Getting the appropriate RGB/GRB color value using the determined color scheme and brightness.

Other changes:
1. Some variables names were changed to be more descriptive, such as the effects volume.
2. There is now an audio feedback when you've reached the upper or lower bounds of the dimming options, making it possible to know whether you've reached the correct state without looking at the pack (like while wearing it).
3. Similarly, the volume controls now emit a feedback sound to indicate when most reach the respective min or max settings (effects, music, overall).


gpstar81:
Overall it looks good. I had a issue with the PowerCell always being red and there was a bug (from my previous code) where updating the colours would turn on led #1 in the Cyclotron if you switched too fast.

I fixed these bugs, refactored the code a bit and cleaned up the function prototype compile warnings.

I also added some more wand synchronisation settings with the pack when you plug a wand in while the pack is running.

Give things a try and if it's good for you, then we can merge it.

DustinGrau:
Additional changes are good, and just tested on my pack through all themes and firing modes. If you can approve the PR, you are welcomed to merge.

gpstar81 (approved):
Looks all good.

PR #118, opened 2023-07-27
EEPROM configuration menu system

gpstar81:
A second EEPROM configuration menu system for adjusting Neutrona and Pack default settings.


source/NeutronaWand/NeutronaWand.ino
```diff
@@ -5799,6 +5917,112 @@ void wandSerialSend(int i_message) {
...
+    for(unsigned int i = 0 ; i < sizeof(objEEPROM); i++) {
+      EEPROM.put(i, 0);
+    }
+
+    updateCRCEEPROM();
+  }
+
+  void saveEEPROM() {
+    // (Video Game Modes) + Cross The Streams / Cross The Streams Mix / Over Heating    
+    uint8_t i_cross_the_streams = 1;
```
DustinGrau, line 5932: Would there be any benefit in just having a singular variable for the "firing mode" which is a byte to represent the current 3 modes and any future modes? eg. VG mode = 1, CTS = 2, CTS_Mix = 3, etc. rather than tracking 2 or more separate values? The available options could then be handled as ENUM values which I believe are bytes or at least unsigned int's? That way if you happen to add more options to this space later you're still only reading/writing a single value.

source/ProtonPack/ProtonPack.ino
```diff
@@ -4663,13 +4683,79 @@ void readEEPROM() {
...
+
+  if(b_cyclotron_single_led != true) {
+    i_single_led = 1;
+  }
+
+  if(b_stream_effects != true) {
+    i_proton_stream_effects = 1;
+  }
+
+  unsigned int i_eepromConfigAddress = EEPROM.length() / 2;
```
DustinGrau, line 4741: Is this just looking for the midpoint of the available EEPROM space and getting the starting byte of memory to use? Do we know this would always be an even address to avoid an error? What if there are more EEPROM settings later, how might you select the starting point? It feels like it may be more safe to carve out more distinct values, leaving buffer in your first space for the LED values to avoid potential overlap, rather than just an arbitrary point in the memory space.

DustinGrau (changes requested):
Looking for some minor changes to naming to avoid potential confusion later, but also looking for clarity with some questions related to the EEPROM space.

source/ProtonPack/ProtonPack.ino
gpstar81: This is the midway point in the entire EEPROM buffer. I figured the first half of the buffer for the LED stuff and start at the midpoint for the configuration settings. There is so much buffer space, we are barely using any of it. Lots of space available.

source/NeutronaWand/NeutronaWand.ino
gpstar81: Hmm perhaps, but cross the streams and mix versions are user configurable editable values, so it is easier for people to set it to true or false if they are manually flashing the boards vs having people mess around with enums and or using more memory by using both enums and the booleans.

source/ProtonPack/ProtonPack.ino
```diff
@@ -4663,13 +4683,79 @@ void readEEPROM() {
...
+    i_single_led
+  };
+
+  // Save to the EEPROM.
+  EEPROM.put(i_eepromConfigAddress, obj_eeprom);
+
+  updateCRCEEPROM();
 }
 
 void clearEEPROM() {
```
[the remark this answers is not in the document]
gpstar81: Ok will rename.

source/ProtonPack/ProtonPack.ino
```diff
@@ -4663,13 +4683,79 @@ void readEEPROM() {
...
   // Clear out the EEPROM only in the memory addresses used for our EEPROM data object.
   for(unsigned int i = 0 ; i < sizeof(objEEPROM); i++) {
-    EEPROM.write(i, 0);
+    //EEPROM.write(i, 0);
```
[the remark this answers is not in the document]
gpstar81: oops

source/ProtonPack/ProtonPack.ino
DustinGrau: I see, they say there's 1024 bytes available, which for this purpose is indeed plenty. I'm just wondering if there could be more features you may add later that may somehow require a larger space, and now there's less than 512 bytes of contiguous space available anywhere. If you reserved 64 bytes for the LED and 64 for the config, you've still got 896 bytes available if you need to store something bigger.

source/NeutronaWand/NeutronaWand.ino
DustinGrau: With your inclusion of the EEPROM menus, you've effectively made the user-configurable values a feature exclusive for advanced users--those who wouldn't understand an ENUM no longer need to make those changes. It's probably more of a line in the sand to be drawn for what type of users you expect to be digging into the software vs. ease of maintenance.

DustinGrau (commented):
Renames are good, just pushing through an updated review so my other comments appear.

source/ProtonPack/ProtonPack.ino
gpstar81: Actually there is 4096 bytes on the chips being used in the Proton Pack and Neutrona Wand :)

DustinGrau: Well then, I read the wrong spec sheet!

ADVCONFIG.md
```diff
@@ -81,7 +81,8 @@ Refer to the [Operation Manual](OPERATION.md) for more information on how to ent
...
+| const bool b&#95;eeprom | Set to true if you to read the configuration data from the EEPROM memory in the Proton Pack.<br><br>`The EEPROM memory has a specified life of 100,000 write/erase cycles. You will generally not reach this limit in your lifetime, but if you did, it can be disabled to bypass it.` | true |
+| const bool b&#95;cyclotron&#95;haslab&#95;chsv&#95;colour&#95;change | When using 12 LEDs for the Cycloton Lid (Stock Haslab), the system will default it to always red because changing the colour space with a CHSV colour object affects the brightness slightly for non RGB pixels such as the ones used in the Haslab Cyclotron Lid. If you are using 12 LEDs that are RGB, they will always be red when this is set to false. Setting this to true will override it and make the LEDs change colour if they are RGB.<br><br>`This only affects when you are using 12 LEDs for the Cyclotron Lid (Haslab)` | false |
```
DustinGrau, line 85: Wow, did not realize that had an effect on the Haslab stock lights, but glad you caught this. Arguably the upgrade options available would be either 20 or 40, so the 12 would have to be a custom solution if someone had RGB lights.

source/NeutronaWand/NeutronaWand.ino
DustinGrau: The more I've thought about this, It's a non-issue and has a larger footprint to change the flags involved. Consider this resolved.

source/ProtonPack/ProtonPack.ino
```diff
@@ -4663,25 +4683,90 @@ void readEEPROM() {
...
+      else {
+        b_cyclotron_single_led = false;
+      }
+    }
+  }
+}
+
+void clearConfigEEPROM() {
+  // Clear out the EEPROM data for the configuration settings only.
+  unsigned int i_eepromConfigAddress = EEPROM.length() / 2;
```
DustinGrau, line 4715: I kept thinking back to this all of last night. As you had pointed out, the size of the EEPROM is perfectly large enough to fit the data to be stored plus extra for future use, but the calculation feels arbitrary. My main concern is how do you choose a location should a 3rd EEPROM space be needed? Would that use a different calculation? If its decided that these locations must change (or do change somehow) it obviously throws off the stored values. The more long-term approach that feels right is picking a distinct base-2 value (32, 64, 128, etc) and using that as the marker for this secondary block of values.

DustinGrau (changes requested):
Overall good, and ready to approve. Though revisiting some thoughts from yesterday I think my main concern is around the new EEPROM address picked for the config data in the pack using a calculation rather than a distinct value. Something about that makes me feel uneasy given that this is meant to store user data, and thus makes the chosen value permanent moving forward.

gpstar81:
> Overall good, and ready to approve. Though revisiting some thoughts from yesterday I think my main concern is around the new EEPROM address picked for the config data in the pack using a calculation rather than a distinct value. Something about that makes me feel uneasy given that this is meant to store user data, and thus makes the chosen value permanent moving forward.

If any more user data gets saved to the EEPROM, the 2 objects would just get enlarged. For example say there was a led addon which had custom values, the led object can have another entry and it gets appended to it in the space where the LED configuration is written to. The same for user configurations on pack operation, enlarge and append to the end of the config object. So the first half of the EEPROM is reserved for non pack like operations and the last half is reserved for operation settings.

Worst case scenario is code could be written to read the existing data and resave it into a new sequence in the EEPEOM address space, but I can't envision that ever needing to happen since such a small amount of data is ever written to the EEPROM and it's not something I expect to get used a lot.

DustinGrau (approved):
The last comment added the clarity needed. Now I understand that the plan is to only have 2 "buckets" for EEPROM data, at the start and middle of the storage area, for either device or configuration options. If any further changes are needed to either, those objects will grow but will continue to be stored at one of those 2 locations.

PR #123, opened 2023-07-30
Doc updates

DustinGrau:
Please take a look at the details I'm now providing on the PCB pages as I started doing more on a basic hookup guide. I appreciate your fixes to my labels/mixups on the smoke diagrams, and hopefully I've not repeated those mistakes here.


PACK_PCB.md
```diff
@@ -37,8 +37,8 @@ Connections for the pack should be made according to the table below.
 
 For connecting the UART pins, use a suitable FTDI chip such as the same **FTDI Basic 5V** chip used for programming the WavTrigger. Connect each pin to the respective label to the FTDI chip.
 
-- If using an FTDI to USB cable, the RTS pin will connect to the DTR board pin.
-- The CTS pin on any FTDI device will not be used (connects to GND on the board).
+- If using an FTDI to USB cable, the RTS pin will connect to the DTR pin on the PCB.
+- The CTS pin on any FTDI device will not be used, but must connect to a GND on the PCB.
```
gpstar81, line 41: Technically the CTS doesn’t need to be connected but it can be.

gpstar81:
Looks good except the wand diagram is using the older 1.0 and not the newer 1.1.

Do you think also the original unmarked diagrams should be included somewhere on the pages as well, so people can see the markings on the boards better?

PACK_PCB.md
DustinGrau: I was correcting it to read that the connection still needed to be made, as I didn't want people to think they had to pull the pin from the connector (and in fact that led to my near-overheat issue originally). Should I just remove that part entirely and just say to connect all wires as directed by the manufacturer since you match the expected pins already?

DustinGrau:
> Looks good except the wand diagram is using the older 1.0 and not the newer 1.1.
> 
> Do you think also the original unmarked diagrams should be included somewhere on the pages as well, so people can see the markings on the boards better?

I did not realize I was out of date on the wand diagram. Do you want to email me a more recent copy and I can update?

PACK_PCB.md
gpstar81: Hmm maybe perhaps indicate they should pay extra attention to the ftdi cable and to not plug it in reverse, that the colours on the cable connectors match the labelled uart header pins. Perhaps a image of a ftdi (or even take a image of a cable plugged into the boards) and to make sure the red 5v is lined from the cable to the vcc pin on the uart headers? I will e-mail over some additional images.

gpstar81:
> > Looks good except the wand diagram is using the older 1.0 and not the newer 1.1.
> > Do you think also the original unmarked diagrams should be included somewhere on the pages as well, so people can see the markings on the boards better?
> 
> I did not realize I was out of date on the wand diagram. Do you want to email me a more recent copy and I can update? EDIT: I see, you already updated the image. I'll fix.
> 
> I can also include the unmarked boards in the detailed sections below each of the diagrams.

Sure thing, I will email the latest over.

gpstar81 (approved):

PR #124, opened 2023-07-31
Updated the Afterlife Wand sound effects and added the ability for the Pack to play them back.

gpstar81:
Updated the Afterlife Wand sound effects and added the ability for the Pack to play them back.


DustinGrau (approved):
If you're feeling like the new changes are good, I've just put them into play and didn't see any adverse effects from my light testing. The new sounds are great, and having the default/option for the wand-to-pack sounds is good. I may revisit the timings on the cyclotron at a later time once the Frutto replacement LED's are in my hands.

PR #126, opened 2023-08-03
Speaker Update

DustinGrau:
I've tried out a new speaker which is super-lightweight but also meets the diameter, wattage, and frequency response equal (or better) to the pair we originally had linked. I've modified the top brackets to make it possible to mount this model. It's a tight fit to get the mount in first and secured, then add the 4 screws around the perimeter--the top is easy but the bottom support needed some mods to better access with tools. https://a.co/d/2mHx8KR


stl/pack/Speaker Mount V2/Top_Speaker_Mount.stl
gpstar81, line 1: Does this alter the shape in any form from using the existing speakers? Perhaps it should go in a Speaker Mount V3 folder or a different name? I say his because we have been mass 3d printing the existing bracket for the speaker support kits and do not want to confuse the previous version to this one.

stl/pack/Speaker Mount V2/Top_Speaker_Mount_Bracket.stl
gpstar81, line 1: Just holes added only or was there shape differences altered? Maybe like the modified Top Speaker Mount, put in a V3 folder or have a different name? As we have been mass 3D printing the current bracket.

gpstar81 (changes requested):
We have been mass 3D printing the current brackets, maybe perhaps give the updated ones a V3 name or put in a V3 folder so we do not get the old brackets mixed up. Are there any substantial changes to the shape of the brackets that could affect the old speaker setups? or just added holes only for mounting?

DustinGrau:
For sake of closing the loop here, I reverted the original top bracket and added the new file with a "_Square" suffix to denote it's use. The supporting top mount is the same new version, as it only removed a bit of unused material for the sake of better tool access to the speaker mounting screws.

PR #137, opened 2023-08-11
eeprom spectral mode

gpstar81:


DustinGrau (approved):
All good, preferences now survive a full power-off and are restored

PR #138, opened 2023-08-12
Spectral custom mode

gpstar81:
-Spectral custom mode. Default colour is purple, but the colours can be controlled and saved from the EEPROM LED Menu system. Refer to the operational guide for usage.
-Updated Spectral Rainbow and Holiday firing so it looks more like it is shooting rainbows or holiday colours from the Neutrona Wand barrel.


DustinGrau (approved):
Amazing additions! Everything makes sense here, and the demo video shows everything quite clearly. I'll merge then test it out on my pack.

PR #139, opened 2023-08-15
desaturate to white when reaching red

gpstar81:


DustinGrau:
I was just looking at this in my own code and was about to try out a quick fix. Instead of the saturation setting, I just make color value 254 be white (hue: 100, saturation: 0) instead of a separate variable. How does the saturation setting work from the EEPROM menu, do you just keep turning the dial to make the red color lighter?

DustinGrau:
Code is loaded on my pack and wand but I'm not sure if I've missed something. Turning the dial on menu 4 (for wand) doesn't seem to do anything once I reach red, whether clockwise or counter-clockwise. Does the effect only work in one direction?

DustinGrau:
I got the wand working--must not have turned the knob enough to make a noticeable change at first. Though think something might be wrong with the inner cyclotron as I saw my colors suddenly change to a desaturated value. I was turning clockwise to go back to purple and somewhere in the middle of the transition I saw it blink and suddenly go a much lighter color. No matter how much I turn I can't get my color back. Perhaps a serial communication glitch reset a value to make it think it should begin altering the saturation instead of the hue?

gpstar81:
> I got the wand working--must not have turned the knob enough to make a noticeable change at first. Though think something might be wrong with the inner cyclotron as I saw my colors suddenly change to a desaturated value. I was turning clockwise to go back to purple and somewhere in the middle of the transition I saw it blink and suddenly go a much lighter color. No matter how much I turn I can't get my color back. Perhaps a serial communication glitch reset a value to make it think it should begin altering the saturation instead of the hue?

Ooops. Pull the latest bug fix and reflash again, it should work.

DustinGrau:
Good catch, that's got it. I ran the full spectrum and saturation for the inner cyclotron and it works now.

DustinGrau (approved):
Tested everything out, seems to do what is expected (only desaturates on counter-clockwise, per clarifications).

PR #140, opened 2023-08-20
demo light mode, auto startup

gpstar81:
Advanced configuration toggle, it makes the Proton Pack turn on automatically after receiving power.


DustinGrau (approved):
Changes make perfect sense. It may need to be noted to the user who requested this that the feature only affects the pack. I don't know if they also intended for the wand to be active as well.

PR #146, opened 2023-08-23
updated afterlife and level 5 firing sounds and mixing

gpstar81:


DustinGrau:
Thank you so much for this! I was nearly there with my attempts at the code, but may have been stymied by not having the right audio file to notice the changes. I did make a single commit where I had found some GB2 firing sounds should have instead been GB1, and what looked like duplicate calls to stop an effect.

DustinGrau (approved):
Looks good, and functions as expected. Thank you!

PR #149, opened 2023-09-05
Early Attenuator integration, music checking switched to the proton pack, wt firmware update, wt library integration

gpstar81:
Early Attenuator integration.
Music checking switched to the proton pack. Music will keep on playing even when a wand is no longer connected. 
wt firmware update, version 1.40 can be found in the extra folder. Please update all your wav triggers with this new firmware.
wt library integration


DustinGrau (approved):
All changes were reviewed yesterday, and uploaded to my pack and wand for testing. No issues encountered.

PR #150, opened 2023-09-08
Ledw

gpstar81:
LED-W control separate from the n-filter jewel.  It will not strobe but instead stay on full during any venting sequences.

Toggle flag request for the 28 segment bargraph so Afterlife mode and any future modes will make it act like 1984/1989 ramping up/down.


DustinGrau (approved):
Changes look sane, no concerns. For reference I've tested the LED-W updates but not the bargraph changes.

PR #153, opened 2023-09-10
Fix compilation errors

nomakewan:
This PR resolves the compilation errors displayed in the Arduino IDE by more strictly enforcing datatypes rather than leaving conversions and comparisons up to the compiler.

### DIFF
--- a/source/NeutronaWand/Header.h
+++ b/source/NeutronaWand/Header.h
@@ -107,7 +107,7 @@ millisDelay ms_switch_mode_debounce;
 millisDelay ms_switch_mode_firing;
 millisDelay ms_intensify_timer;
 millisDelay ms_firing_debounce;
-const unsigned int i_firing_debounce = 500;
+const unsigned int i_firing_debounce = 50;
 const unsigned int i_intensify_delay = 400;
 const uint8_t i_switch_mode_value = 200;
 const uint8_t i_switch_barrel_value = 100;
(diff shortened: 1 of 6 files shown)


DustinGrau:
While I'm good with the changes to the true "gpstar" code I would be hesitant to approve the 3rd-party code that was added. That said, I do understand the changed: I can see there's removal of a variable that's written but never read, and casting of a value explicitly rather than implicitly.

nomakewan:
That's understandable; I noticed that when it was added to this project, the headers were modified so that they would work specifically with this project (that is, the user no longer has to edit the header in order to make it work for the gpstar/Mega/Nano setups).

I'd be more than happy to open a PR for those changes on the source repo as well. In fact, I think I'll do just that.

DustinGrau:
For reference, is this the upstream repo you mentioned?
https://github.com/robertsonics/WAV-Trigger-Arduino-Serial-Library

nomakewan:
Yep! That's the one. I have an active PR there now to fix several issues, which I have also replicated here.

nomakewan:
Looks like the latest commit added a unique serialFlush() function to the Proton Pack copy of WavTrigger, so just to be sure I've updated the branch of this PR. Looks like it's still good to go.

DustinGrau:
Michael was experimenting with some additional changes which may or may not move forward. I think he plans some cleanup. We both have some travel/plans lined up so the software will go on a brief break before that occurs.

nomakewan:
I've merged main into this so that the newly-instituted compile checks would run on the PR so that it could more easily be approved for merging. The other PR has likewise been updated to account for the new checks. Apologies for the delay in updating, I was away from my lab.

DustinGrau:
No apologies needed, I had not expected Michael to approve my changes early. I had just returned from my trip overseas and he is still on vacation most of this week. This was something we had spoken about implementing, though ideally we want to pre-build the binary firmware files for easy uploading as the end result. I happened to incorporate some of your changes as part of my testing the compilation process--I'm holding out on the WavTrigger updates until Michael returns and implements some items he's been thinking over.

nomakewan:
Gotcha. I'll keep an eye out for the changes and update as necessary.

PR #154, opened 2023-09-10
Various fixes

gpstar81:
Debounce error mode on the wand if you try to button smash firing or CTS firing, which helps prevent unexpected crashes on the Proton Pack.


DustinGrau (approved):
New behavior should be a good deterrent to excessive and unusual behavior by users. The goal is to penalize the mashing of buttons during any firing sequence, while allowing the system time to recover from the resulting flurry of serial commands.

DustinGrau:
I have a branch open for updating documentation and I see a pending PR for some other doc changes. I can plan to add something in my branch which addresses the error mode from this new behavior there.

PR #156, opened 2023-09-12
fade out cyclotron when shutting down

gpstar81:


source/ProtonPack/ProtonPack.ino
```diff
@@ -4561,24 +4608,24 @@ void checkWand() {
 
               switch(i_mode_year) {
                 case 2021:
-                  stopEffect(S_AFTERLIFE_CROSS_THE_STREAMS_END);
-                  stopEffect(S_AFTERLIFE_CROSS_THE_STREAMS_START);
+                  //stopEffect(S_AFTERLIFE_CROSS_THE_STREAMS_END);
```
DustinGrau, line 4611: Were these left commented out for testing, or to eventually be removed>

source/ProtonPack/ProtonPack.ino
```diff
@@ -226,6 +236,7 @@ void loop() {
         b_2021_ramp_up = false;
         b_2021_ramp_up_start = false;
         b_inner_ramp_up = false;
+        b_fade_out = true;
```
DustinGrau, line 239: This only appears to affect the 2021 mode, was that correct? I did notice it during the shutdown but wasn't sure if there was an intended impact to the 84/89 modes as I didn't notice anything different.

source/ProtonPack/ProtonPack.ino
```diff
@@ -3031,27 +3070,18 @@ void wandFiring() {
...
-  ms_cyclotron_auto_speed_timer.stop();
-
-  // Adjust the gain with the Afterlife idling track.
-  if(i_mode_year == 2021 && i_wand_power_level < 5) {
-    if(ms_idle_fire_fade.remaining() < 1000) {
-      adjustGainEffect(S_AFTERLIFE_PACK_IDLE_LOOP, i_volume_effects, true, 30);
-    }
-    else {
-      adjustGainEffect(S_AFTERLIFE_PACK_IDLE_LOOP, i_volume_effects, true, ms_idle_fire_fade.remaining());
+void modeFireStopSounds() {
```
DustinGrau, line 3073: I presume this allows for a small delay to the wavTrigger so that the communications from the wand and pack can have time to be received and take action?

source/ProtonPack/ProtonPack.ino
```diff
@@ -3114,77 +3158,77 @@ void wandStoppedFiring() {
...
 
-      case SLIME:
-        stopEffect(S_SLIME_START);
-        stopEffect(S_SLIME_LOOP);
-        stopEffect(S_SLIME_END);
-      break;
+      stopEffect(S_FIRING_LOOP_GB1);
+      stopEffect(S_GB1_FIRE_START_HIGH_POWER);
+      stopEffect(S_GB1_FIRE_HIGH_POWER_LOOP);
+      //stopEffect(S_FIRE_START_SPARK);
```
DustinGrau, line 3183: Again, just checking if commented for testing or to be removed?

DustinGrau (commented):
No issues seen while testing, but had some questions on what impact some changes have on operation and whether certain commented code was for testing purposes only or earmarked items to be removed from the normal chain of actions.

source/ProtonPack/ProtonPack.ino
gpstar81: Correct, 2021

source/ProtonPack/ProtonPack.ino
gpstar81: So I found during testing if I spread out several of the serial commands with tiny delays, it makes the system more stable during the fastled updates during any button smashing.

source/ProtonPack/ProtonPack.ino
gpstar81: At some point I'd like to re implement but because of the wand overload feature they are no longer critical. I was lessening the burden on the serial buffer queue. Particularly on sound effects which are very short in nature and can play out instead of sending a serial command to stop it.

source/ProtonPack/ProtonPack.ino
gpstar81: As mentioned in a previous comment, I was lessening the burden on the serial buffer queue. Particularly on sound effects which are very short in nature and can play out instead of sending a serial command to stop it. For now I would leave it commented out but in the code.

source/ProtonPack/ProtonPack.ino
```diff
@@ -3087,7 +3237,28 @@ void wandStoppedFiring() {
...
+void wandStoppedFiring() {
+  resetFastLed();
+
+  modeFireStopSounds();
+
+  // A tiny ramp down delay helps with the sounds.
+  ms_firing_stop_sound_delay.start(i_fire_stop_sound_delay);
+
+  // Keep this code for later future updates, when using 4 wire led chipsets, the delay timer above can be avoided.
+  //wandStopFiringSounds();
```
DustinGrau, line 3254: Ah, now I think I understand--this _might_ be the reason why I would sometimes still hear effects such as the slime loop still playing if I timed my actions right, as this code would not execute at this point. Though I see there's a timer which is checked in the debounceChecks() which will call this from the main loop. I'll try to note if I come across that issue again as it may make sense to re-enable to ensure the looping effects get stopped.

DustinGrau (approved):
Casting my approval on the latest changes. The new error mode lock-out for abusive users, and the compromise on prioritizing serial communications over the LED updates offers more protection from incorrect usage.

PR #158, opened 2023-09-14
Stab

gpstar81:
reverting back many changes:


DustinGrau (approved):
As it currently exist, the changes were working well for me and didn't cause any unexpected issues. Normal use and even aggressive button actions were caught by the new lock-out feature on the wand. Per outside discussion many of the changes to the LED timings and serial communications were rolled back.

PR #160, opened 2023-09-15
Wand firing debounce

gpstar81:
Wand firing debounce to help with button smashing.


DustinGrau (approved):
Good fix for the user input. No apparent alteration to responsiveness with normal usage, and it was more difficult to even put the wand into the lockout mode through aggressive button mashing.

PR #161, opened 2023-09-24
Auto-compile test actions

DustinGrau:
Triggers a compilation of all Arduino code on push or creation of a PR as a sanity test. This will provide additional checks which MUST pass in order for the code to be deemed ready for merging. This should be in addition to a proper code review, but at least confirms that compilation is not broken by any changes.

The current demonstration of the workflow will cause a compilation of the "core" code using the Arduino Mega platform, while the "addons" will be compiled using the Arduino Nano platform.


gpstar81 (approved):

DustinGrau:
Thank you Michael. I was going to tag you for review when I thought you might be back from vacation.

gpstar81:
Yes, back on Thursday. Lots of things on my agenda to do when getting back...

PR #162, opened 2023-09-27
heatsink and speaker insert stl's

gpstar81:
Heatsink and speaker insert stl's from Frutto Technology for the Neutrona Wand.


DustinGrau (approved):
Nice, Github has a 3D viewer for STL files in the browser. That's convenient!

PR #164, opened 2023-09-29
Fixes button smashing crash

gpstar81:
Additional data checking on the wav trigger serial communication to prevent the system from crashing during track reporting while button smashing the Neutrona Wand.


DustinGrau (approved):
Code looks good, only pushed in a whitespace fix (the wavtrigger files used tabs instead of spaces). Everything has been tested on my pack/wand with perfect results. No issues experienced in any modes, even with button smashing. If this is good with you, we can move forward with a merge and call this bug squashed!

PR #167, opened 2023-09-30
flasher software and precompiled binaries

gpstar81:
A simple flasher software and precompiled binaries.


DustinGrau (commented):
Content is perfectly fine, as is all new folder/artifacts. Unfortunately the flashing utility is not working for me on my Windows10 laptop. It found a COM3 device which was correct but would not upload because "target is not responding".

nomakewan:
Testing the software on my Windows 7 dev rig, it does come up, though it initially tries COM1 (which doesn't exist) but then once clicked locks onto COM12 (which is correct).

I did not test flashing however since I am not so sure about the precompiled binary and wanted to make sure there was clarification before flashing. I normally just would update the GPstar board using the Arduino IDE and just uploading the sketch. However doing that assumes there is a bootloader, and since it works I have to assume that there is indeed a bootloader in place on the GPstar board. Comparing the filesize of the precompiled binaries in this PR to those generated by my IDE, it appears the binaries here are the "no bootloader" version.

To flash raw hex to the chip would require more than just your standard 5V UART cable, wouldn't it? Normally that would be done with ArduinoISP or similar device. And if you flash a non-bootloader firmware over a bootloader firmware, it will no longer accept sketches via the Arduino IDE as the GPstar currently does.

Does this sound correct, or am I missing something?

gpstar81:
> Testing the software on my Windows 7 dev rig, it does come up, though it initially tries COM1 (which doesn't exist) but then after selecting a firmware and clicking on the COM port drop-down it locks onto COM12 (which is correct).
> 
> I did not test flashing however since I am not so sure about the precompiled binary and wanted to make sure there was clarification before flashing. I normally just would update the GPstar board using the Arduino IDE and just uploading the sketch. However doing that assumes there is a bootloader, and since it works I have to assume that there is indeed a bootloader in place on the GPstar board. Comparing the filesize of the precompiled binaries in this PR to those generated by my IDE, it appears the binaries here are the "no bootloader" version.
> 
> To flash raw hex to the chip would require more than just your standard 5V UART cable, wouldn't it? Normally that would be done with ArduinoISP or similar device. And if you flash a non-bootloader firmware over a bootloader firmware, it will no longer accept sketches via the Arduino IDE as the GPstar currently does.
> 
> Does this sound correct, or am I missing something?

These binaries are the non bootloader ones. The gpstar boards have the bootloaders on them already, this does not overwrite the bootloader when flashing the non bootloader version. Jonathan just tested a updated flasher and it works for him. There is a baud rate drop down now, it should be default 115200 which is required for flashing these chips.

I have only tested this on vritual machines but it works no problem for me. I do find on a fresh install of Windows, plugging in the cable, it takes windows a few minutes to setup the USB TTL device, then it can detect the board after and flash no problem.

nomakewan:
> > Testing the software on my Windows 7 dev rig, it does come up, though it initially tries COM1 (which doesn't exist) but then after selecting a firmware and clicking on the COM port drop-down it locks onto COM12 (which is correct).
> > I did not test flashing however since I am not so sure about the precompiled binary and wanted to make sure there was clarification before flashing. I normally just would update the GPstar board using the Arduino IDE and just uploading the sketch. However doing that assumes there is a bootloader, and since it works I have to assume that there is indeed a bootloader in place on the GPstar board. Comparing the filesize of the precompiled binaries in this PR to those generated by my IDE, it appears the binaries here are the "no bootloader" version.
> > To flash raw hex to the chip would require more than just your standard 5V UART cable, wouldn't it? Normally that would be done with ArduinoISP or similar device. And if you flash a non-bootloader firmware over a bootloader firmware, it will no longer accept sketches via the Arduino IDE as the GPstar currently does.
> > Does this sound correct, or am I missing something?
> 
> These binaries are the non bootloader ones. The gpstar boards have the bootloaders on them already, this does not overwrite the bootloader when flashing the non bootloader version. Jonathan just tested a updated flasher and it works for him. There is a baud rate drop down now, it should be default 115200 which is required for flashing these chips.
> 
> I have only tested this on vritual machines but it works no problem for me. I do find on a fresh install of Windows, plugging in the cable, it takes windows a few minutes to setup the USB TTL device, then it can detect the board after and flash no problem.

Thank you for the clarification! I just tested it and it worked perfectly. Great work!

DustinGrau (approved):
I have just re-tested with the latest flasher and these are the results:

-Proton Pack: Using the 5-pin USB cable which is supplied with the kit for flashing works great. The correct COM port is found (must be selected for use) and the BAUD rate is set appropriately. My pack flashed with the precompiled HEX file just fine.

-Neutrona Wand: This did not work but I think I know why. In my case I have hardwired the 5V FTDI chip to my board, the same chip which is used to program the WavTrigger. This device's COM port is found properly, but the flashing operation does not work as expected. This may be the difference between the two setups and why I only seemed to have an issue.

Given that the "stock" cable provided works for myself and others, I'm going to say this is suitable for most end-users as they would only need to flash new software when absolutely needed. For power users such as myself, I have means of using the Arduino IDE directly which works just fine with all my devices.

tl;dr: This is more than enough for the populace, let's ship it.

PR #168, opened 2023-10-02
Condensed code for diy Neutrona Wands

gpstar81:
Condensed code for diy Neutrona Wands.

Updated wav trigger firmware. It is not required to update to this if you are already running 1.40 firmware. This was just a minor fix I requested.


DustinGrau:
Is it possible to add some instructions on how to generate or obtain the .hex files from a compile for inclusion with each pull request? Given that every code change will need a recompile, it may be a new code review item to ensure a new .hex is also provided. Likewise, we should probably remove the "Vx.x.x" subfolder so that each merge to the main branch is always the latest--if someone needs an earlier version they should use the TAGS link to pull binaries from an earlier release. Just a thought to make that portion easier to keep up with.

gpstar81:
> Is it possible to add some instructions on how to generate or obtain the .hex files from a compile for inclusion with each pull request? Given that every code change will need a recompile, it may be a new code review item to ensure a new .hex is also provided. Likewise, we should probably remove the "Vx.x.x" subfolder so that each merge to the main branch is always the latest--if someone needs an earlier version they should use the TAGS link to pull binaries from an earlier release. Just a thought to make that portion easier to keep up with.

That should be no problem, just need to look into it. Just crazy busy with getting a Mac flasher ready, which is almost there, plus all the other things being discussed etc.

DustinGrau:
Not a worry, I just found a way using the arduino-cli on my Mac installation, which seems to work exactly as intended. The compiler is sort of hidden away in the .github folder as I'm hoping this might be reusable on the Github workflows at some point. I've restructured the binaries folder to hold the various HEX files for specific AVR platforms and customizations.

DustinGrau (approved):
Original code changes were reviewed and look good. I've not yet had a chance to test on my devices, but I also don't want to hold up release as I'm sure Michael has done the necessary testing. 

One thing I changed in this PR was to add a new compile.sh script which can compile all of the devices and save their non-bootloader .hex files to the binaries folder. I've done away with the version numbers to avoid confusion, as versioning is a function of the Github repo and shouldn't be part of the build process. This will allow us to build new binaries whenever code is submitted for review, and the script will produce for both the Mega and Nano AVR platforms while also generating custom binaries where necessary by changing specific config options.

PR #172, opened 2023-10-06
Firmware Flasher for Macs

gpstar81:


DustinGrau (approved):
Beautiful! Works on both Intel and M1 architectures for Mac, and flashed both the pack and wand flawlessly when using the standard USB programming cable. This time I didn't try with my special FTDI chip in the wand but with the spare boards for the wand and pack.

DustinGrau:
Made some updates to doc, adding my latest video for the Attenuator and clarifying file extensions on the flasher programs. Also stated Intel/M1 for the Mac application.

PR #173, opened 2023-10-06
dmg of mac flasher

gpstar81:


DustinGrau (approved):
New installer works to hide the contents of the .app file, but requires some adjustments to the documentation. Also, recent changes to the binaries required and update to a screenshot, so I've done both changes.

PR #181, opened 2023-10-10
Update CTS Mix Functionality

nomakewan:
This proposed change makes CTS Mix into an independent optional fire mode that can be used alongside the video game or spectral fire modes.

In this version, CTS Mix works only when the user is firing a Proton Stream and then holds down the Barrel Wing Button. If they release it, it returns to Proton Stream as before. If they release Intensify, the wand stops firing even if the Barrel Wing Button remains held. When not firing, the Barrel Wing Button continues to be a mode switch as if CTS were not enabled.


gpstar81 (approved):
Great idea to bring a variation of CTS into the default video game mode.

DustinGrau:
Oh this is perfect. I was wondering about being able to switch up to the CTS mode on the fly while in the VG mode. I was about to start researching whether we could add an API call to switch it on via the Attenuator. I like this much better.

PR #183, opened 2023-10-11
EEPROM Config Sub Menu + other updates

gpstar81:
-EEPROM Config Sub menu with more options that can be saved and set (refer to the operational guide manual)
-CTS Mix is now enabled in Video Game Mode.
-Various compiler error fixes and other minor bug fixes.
-Updated binaries, custom binaries are now removed.

*** Make sure to update the sound files on your SD card***


nomakewan:
It appears to actually be CTS that is enabled in video game mode, not CTS Mix. When I'm in video game mode and hold the wing button it plays the CTS sound, but when I release it doesn't end CTS, it just keeps firing. Then when I release intensify, it plays all the CTS end sounds. CTS Mix now appears to be the same as it was before.

To be honest, is there a reason to have CTS/CTS mix? And if removed, would it be better to have the CTS implementation with the VG modes be the way it is here (once you start crossing the streams you must stop firing to uncross them) or the CTS Mix way (you can uncross the streams by releasing the wing button and keep firing a Proton Stream).

I was testing these new changes this morning and am in the middle of updating the documentation to cover the current state of functionality, but I figure I'll write my thoughts now as well while I'm doing that.

gpstar81:
> It appears to actually be CTS that is enabled in video game mode, not CTS Mix. When I'm in video game mode and hold the wing button it plays the CTS sound, but when I release it doesn't end CTS, it just keeps firing. Then when I release intensify, it plays all the CTS end sounds. CTS Mix now appears to be the same as it was before.
> 
> 
> 
> To be honest, is there a reason to have CTS/CTS mix? And if removed, would it be better to have the CTS implementation with the VG modes be the way it is here (once you start crossing the streams you must stop firing to uncross them) or the CTS Mix way (you can uncross the streams by releasing the wing button and keep firing a Proton Stream).
> 
> 
> 
> I was testing these new changes this morning and am in the middle of updating the documentation to cover the current state of functionality, but I figure I'll write my thoughts now as well while I'm doing that.
> 
> 
> 
> Oh, and there's a bug with the wand shutdown where if you shut down the wand multiple times rapidly (any more than once) the shutdown sound will play multiple times rather than only once per switch-off. Figure that should be an easy fix too, might toss it in with the doc updates PR.

So CTS and CTS mix modes yes. Some people do not want video game modes and like the alt firing button for firing. I prefer the mix myself but some requested regular CTS where you only tap to start it.

DustinGrau:
The CTS mix sounds accurate based on how it was described that simply pressing the barrel wing button would initiate the action, but releasing it would not end the firing sequence. 

I was also going to throw in another request for a change since the menus are being adjusted, is it possible to use three elements per menu level as opposed to two? So it would be a three on three off pattern to make the display easier to see.

DustinGrau (approved):
Just uploaded all the new software and audio files, and walked through the new EEPROM menu. This is so nice to have the various toggles at the ready!

Only a few minor typos or inconsistencies within doc which have been adjusted.

PR #189, opened 2023-10-16
Minor Wand changes

nomakewan:
This changes the gpstar build code to use digitalRead for the Barrel Wing Button and Barrel Safety Switch as a Mega 2560 has digital buffers on A6 and A7, unlike the Nano. Additionally this change disables both serial read and serial write if b_gpstar_benchtest is enabled, which should prevent any synchronization errors from the pack if a standalone-configured Wand is connected for some reason.

### DIFF
--- a/source/NeutronaWand/NeutronaWand.ino
+++ b/source/NeutronaWand/NeutronaWand.ino
@@ -5614,7 +5614,7 @@ void checkRotary() {
         case ACTION_EEPROM_MENU:
             // Counter clockwise.
             if(prev_next_code == 0x0b) {
-              if(i_wand_menu == 4 && switch_intensify.getState() == HIGH && analogRead(switch_mode) < i_switch_mode_value) {
+              if(i_wand_menu == 4 && switch_intensify.getState() == HIGH && digitalRead(switch_mode) == LOW) {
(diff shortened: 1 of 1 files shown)


gpstar81 (commented):
Looks ok to me.

DustinGrau:
What about other instances of analogRead(switch_mode) outside of the switchMode() function? Is this something that should be a global change, and remove the i_switch_mode_value and i_switch_barrel_value variables?

nomakewan:
> What about other instances of analogRead(switch_mode) outside of the switchMode() function? Is this something that should be a global change, and remove the i_switch_mode_value and i_switch_barrel_value variables?

Those variables must remain because Nano builds require them. The Nano does not have digital buffers on A6 or A7, which is why this code existed in the first place. The Mega 2560 that the gpstar board is built around however has digital buffers on all its analog pins, and so can use digitalRead instead.

I did miss a few gpstar bits of code elsewhere though, and have pushed a commit just now to add them in. Good catch!

DustinGrau (approved):
Thanks for the additional fixes, that covered what I was concerned with. I've uploaded the code and it looks to be working

PR #191, opened 2023-10-17
Add barrel debounce, Afterlife thwoop sound

nomakewan:
This adds a debounce to the barrel extension switch, makes it so the function to check the barrel state is only called once per loop (and all functions which check barrel state now check a global variable instead), and adds a 'thwoop' sound to Afterlife mode when the barrel is first extended.


gpstar81:
Merging this into a working update branch.

PR #193, opened 2023-10-17
Afterlife Barrel Extension Sound

gpstar81:
-Afterlife barrel extension sound
-Condensed code to fit on the diy ATmega328 powered Neutrona Wands.


gpstar81:
In this update, the barrel extension sound in Afterlife is on as default. I am undecided if it should be on by default or not, mainly because the barrel already makes a real world sound when extending it. Tell me what you think? Or should it be off as default and let people manually toggle it on in custom builds?

Perhaps the sound effect could be edited better to remove any background noise hum, or do you think it is ok as is?

DustinGrau:
The audio file mainly contains the higher click sounds, but wasn't there a "thwoomp" component on the low end? Granted, that may not even be audible on any speakers used in the wand. By chance if you have the original, maybe using a low-pass filter to isolate that portion might work and you could play it back on the pack?

nomakewan:
It definitely sounds weaker with the filter applied, yeah. I honestly wish I had one of those 3W speakers in to see what it'd sound like on that versus the 0.25W stock one. On the stock one the non-filtered barrel sound sounded fine, but maybe the tiny distortion at the low end would've been more noticeable on the 3W speaker. A project for when I toss a bunch of Frutto upgrades into my Wand, I suppose.

With the noise thing, 1984 just had a "clack!" noise when extending the barrel so I agree with not making a sound there and leaving it to just the raw sound of the plastic barrel popping out. But Afterlife's sound was unique and not really replicated by the sound the Hasbro Wand makes by itself, thus the added sound here.

Saw the changes to the code; I wasn't sure if rolling the sounds into the switch check function was kosher so I'd separated them originally. Totally agree with the way it is now though!

DustinGrau:
My turn to debug an issue: I've flashed everything from the branch, uploaded the file, but I'm not hearing the sound. I'm certain I copied the WAV file with the correct prefix to my SD cards and I tried both the binary HEX and direct upload from ArduinoIDE with no luck. The wand definitely knows the barrel is retracted as it won't fire, so it's not a physical issue with the switch. Any ideas?

gpstar81:
> My turn to debug an issue: I've flashed everything from the branch, uploaded the file, but I'm not hearing the sound. I'm certain I copied the WAV file with the correct prefix to my SD cards and I tried both the binary HEX and direct upload from ArduinoIDE with no luck. The wand definitely knows the barrel is retracted as it won't fire, so it's not a physical issue with the switch. Any ideas?

I would wait a bit until my next push to this branch. Lots of changes coming.

DustinGrau:
Understood, though I thought this would be an easy check to see how the new audio effect sounded. I'm just unfortunately not getting the expected result.

gpstar81:
Closing this request for now. Will do another request when i'm finished. still lots of debugging and small bugs to fixes in this. A few more days to go.

PR #205, opened 2023-11-18
Audiofix

gpstar81:


gpstar81:
The EEPROM config menu has been updated to be able to disable and enable the beep loop in Afterlife as requested. (See documentation)

DustinGrau:
I've adjusted the Attenuator code to support the new API calls which deliver the current music track. This is now worked into the web UI to update the dropdown with the current track playing, making it much easier to select a track.

DustinGrau (approved):
Revised functionality for pause/resume now works as expected, and all devices are kept in sync with the commands. Currently the Attenuator will keep the start/stop action via the dial, but the web UI will add the ability to perform the pause/resume along with all other track navigation abilities.

PR #209, opened 2023-11-22
Updated Afterlife sound effects

gpstar81:
-The second background ramp sound effect is added for the Neutrona Wand.
-Updated beep sound effect from top toggle switch.


DustinGrau (approved):
Confirmed as just an audio update at this time

PR #211, opened 2023-11-25
Attenuator Updates

DustinGrau:
Clean up and speed up some of the UI code, while adjusting timings for proper sync with the pack. Also introduced a new web API to restart the attenuator/wireless device in the event a re-sync is needed. Added a longer initial delay for the pack to begin the handshake, ensuring the wand serial connection is established and preferences are obtained from EEPROM.


DustinGrau:
@gpstar81 there is a minor update to the pack code for this, as I found there was indeed need for a bit more of a delay before it seeks out and handshakes initially with the attenuator on serial1.

source/ProtonPack/ProtonPack.ino
```diff
@@ -4301,11 +4308,11 @@ void checkSerial1() {
             break;
 
             case A_SYNC_START:
-              //b_serial1_connected = false;
+              debug("Serial1 Sync Start");
```
gpstar81, line 4311: I would probably remove these debug calls, the strings will probably take up memory even though they are not used.

source/ProtonPack/ProtonPack.ino
```diff
@@ -4301,11 +4308,11 @@ void checkSerial1() {
...
             break;
 
             case A_SYNC_END:
-              //b_serial1_connected = false;
+              debug("Serial1 Sync End");
```
gpstar81, line 4315: Same as above.

source/ProtonPack/ProtonPack.ino
```diff
@@ -549,6 +549,13 @@ void loop() {
   }
 }
 
+void debug(String message) {
```
gpstar81, line 552: Maybe encase the entire debug function around the DEBUG_SEND_TO_CONSOLE to cut down on the memory footprint? Though the compiler may be removing it from the code if it's never called, but just to be safe perhaps....

gpstar81 (changes requested):

source/ProtonPack/ProtonPack.ino
DustinGrau: I'll back this out entirely. I got what I needed from the messages, and 99.9% of the time they're not needed. If we need debugs I'll just stick to adding them when needed.

gpstar81 (approved):

PR #215, opened 2023-11-27
Update wand heatup sound

nomakewan:
Replaces the current Original Mode wand heatup sound with one recorded directly from the sound chip on the Spengler wand.

I have also attached the Spengler heatdown sound to this PR. It was not used in the Original Mode code, but in case it is deemed necessary, here it is.

[wand_heatdown_alt.zip](https://github.com/gpstar81/haslab-proton-pack/files/13480768/wand_heatdown_alt.zip)


gpstar81:
Do you think we can keep the old one, and just add the new one at the end of the file listing. I already took the file and made modifications to it to try to make it more clear. You can cancel this request if you want, I will include the file in my next update.

PR #216, opened 2023-11-27
Allow music from standalone wand

nomakewan:
This restores the ability for a standalone wand to utilize full music playback controls, and also corrects some of the code documentation for the menu system generally.

### DIFF
--- a/source/NeutronaWand/NeutronaWand.ino
+++ b/source/NeutronaWand/NeutronaWand.ino
@@ -1326,8 +1326,8 @@ void mainLoop() {
       settingsBlinkingLights();
 
       switch(i_wand_menu) {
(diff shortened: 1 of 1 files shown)


gpstar81 (commented):
I believe with this the track will play but it will still lack track checking: ie: go to the next track when the current track finishes, etc.

I will add it in a few days for the standalone wand.

nomakewan:
Ah right, I forgot to check if that actually worked since I was focusing on the menu options. Would you prefer I closed this and left it to you?

gpstar81:
> Ah right, I forgot to check if that actually worked since I was focusing on the menu options. Would you prefer I closed this and left it to you?

ya probably best.

PR #220, opened 2023-11-30
Neutrona Wand music control in bench test mode

gpstar81:
I haven't had the time to bench test this yet, so it should also be tested with a Proton Pack as well, and the Neutrona Wand as a stand alone.

Whats new:
-Neutrona Wand music control in bench test mode.
-Another sound effect added for the Neutrona Wand, but it is unused at the moment.


DustinGrau:
Looks good on the surface, but yeah this needs testing and I'll plan to do a sequence with the special standalone wand firmware. I just committed some updates including rebuilding the binaries and some minor doc updates to solve some confusion on software/hardware items that have come up recently.

nomakewan:
I got a little busy this week, but I should be able to do a test with a standalone wand this weekend.

At a quick glance it looks very similar to how I did it on my end (including adding in that check timer start I missed!), with just a few modifications I'd make here and there to some syntax. Nothing breaking though.

I'll be able to review it more thoroughly in the morning (~12 hours). Thank you!

DustinGrau (approved):

DustinGrau (changes requested):
Just found a problem: the wand barrel tip is not illuminating when firing. I've done some tests to make sure it wasn't my hardware (didn't crush a wire or anything like that), but just noticed it wasn't seen when firing.

gpstar81:
> Ohhhh, I think I see why. The strobing of that LED was tied to the bargraph animation for Super Hero Mode. There doesn't seem to be similar logic for Mode Original, though a comment in the code implies this happens when the system is close to overheating.

Oh yes I forgot about adding the strobe to mode original firing. I will add it during the weekend and commit it to this pull request.

DustinGrau:
Just to restore the function and make sure it does something at the moment, I added some very basic code which just illuminates the tip when the last LED in the barrel is lit. This makes functional sense as it would follow the animation for the stock lights. You can choose to keep it or replace that logic, though in looking at the bargraph-dependent animation for Super Hero, this approach may make more sense as the new wand LED's are implemented.

nomakewan (approved):
I've pushed a change that just updates some documentation and formatting related stuff. As I said before from my cursory glance, nothing at all major I noticed, just minor stuff.

Test looks good, I'd say that for the purposes of this PR (restoring music functionality), it's good to go! Perhaps leave the wand tip stuff for a subsequent PR? I can't test that particular change as I'm just using the stock Hasbro 5-LED setup with no additional tip on mine.

gpstar81:
I am just redoing the wiring on my bench test boards, I will give things a good test soon and update the wand flash tip at the same time.

gpstar81:
-Updated the strobe in mode original.
-Added a slightly longer delay on the afterlife beep sounds when changing the power level. If you spam the rotary encoder by rotating it super quick for 10+ seconds before, you could get the audio to crash in the wand (very fringe case use). Now it is stable.

DustinGrau (approved):
Reviewed and tested (bench mode and normal), meets my needs and seems to be working as intended.

PR #221, opened 2023-12-03
updated gethue support with led count, updated stls

gpstar81:
-bench test rig stl's added
-updated wand cage
-get hue changes for the wand to support barrel led count. some early work on the new barrel LEDs coming soon.

### DIFF
--- a/source/NeutronaWand/Colours.h
+++ b/source/NeutronaWand/Colours.h
@@ -238,11 +238,12 @@ CRGB getHueAsGRB(uint8_t i_colour, uint8_t i_brightness = 255) {
 
 CRGB getHueColour(uint8_t i_colour, uint8_t i_num_leds, uint8_t i_brightness = 255) {
   switch(i_num_leds) {
-    case 48:
+    case LEDS_48:
+    case LEDS_60:
       return getHueAsRGB(i_colour, i_brightness, false);
     break;
 
-    case 5:
+    case LEDS_5:
     default:
       return getHueAsRGB(i_colour, i_brightness, true);
     break;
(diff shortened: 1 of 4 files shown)


gpstar81:
Nothing really major, it should not break anything. I just wanted to get these early hooks in for the upcoming 48 + 1 barrel LEDs coming soon and a few new STLs.

source/NeutronaWand/Colours.h
```diff
@@ -235,3 +235,16 @@ CRGB getHueAsGRB(uint8_t i_colour, uint8_t i_brightness = 255) {
   // Forward to getHueAsRGB() with the flag set for GRB color swap.
   return getHueAsRGB(i_colour, i_brightness, true);
 }
+
+CRGB getHueColour(uint8_t i_colour, uint8_t i_num_leds, uint8_t i_brightness = 255) {
+  switch(i_num_leds) {
+    case 48:
```
DustinGrau, line 241: Could we instead use an ENUM for STOCK_BARREL and FRUTTO_BARREL just to avoid specific LED counts? I doubt we'll have many other configurations from those, and it gives a more user-friendly name for identification throughout. Asking as you're not really acting upon the count of the LED's but the type of LED's to change from GRB (stock) to RGB (Frutto).

source/NeutronaWand/NeutronaWand.ino
```diff
@@ -1869,7 +1869,7 @@ void mainLoop() {
   }
 
   if(ms_firing_lights_end.justFinished()) {
-    fireStreamEnd(getHueAsGRB(C_BLACK));
+    fireStreamEnd(getHueColour(C_BLACK, i_num_barrel_leds));
```
DustinGrau, line 1872: Related to the request in Colours.h, if we could use an ENUM to track the type of LED configuration vs. the LED count, that may be easier for maintenance (given that this designation covers both a count of LED's as well as the RGB/GRB status).

DustinGrau (commented):
One request based on the current changes: use of an enum rather than a simple LED count. More details in other comments for 2 files but that's the gist of it.

DustinGrau (approved):
That'll work, thanks!

PR #229, opened 2023-12-13
System power inactivity indicator

gpstar81:
After 1 minute of system inactivity while the Neutrona Wand and Proton Pack are powered down, the LED at the front of the Neutrona Wand body will start to blink after 1 minute to indicate the system still has power. This timer resets after any action is taken with the system.


source/NeutronaWand/NeutronaWand.ino
```diff
@@ -1757,6 +1757,47 @@ void mainLoop() {
...
+      // If the power indicator is enabled. Blink the LED on the Neutrona Wand body next to the clippard valve to indicator the system has battery power.
+      if(b_power_on_indicator == true && WAND_ACTION_STATUS == ACTION_IDLE && (b_pack_on != true || b_no_pack == true)) {
+        if(ms_power_indicator.isRunning() == true && ms_power_indicator.remaining() < 1) {
+          if(ms_power_indicator_blink.isRunning() != true || ms_power_indicator_blink.justFinished()) {
+            ms_power_indicator_blink.start(i_ms_power_indicator_blink);
+          }
+
+          switch(SYSTEM_MODE) {
+            case MODE_ORIGINAL:
+              if(b_pack_ion_arm_switch_on != true) {
```
nomakewan, line 1770: This line should be `if(b_pack_ion_arm_switch_on != true || b_no_pack == true)` as the check on line 190 sets b_pack_ion_arm_switch_on to true if b_gpstar_benchtest is set to true. Therefore there is no way for b_pack_ion_arm_switch_on to ever be false if b_gpstar_benchtest is true.

nomakewan (changes requested):
Only a super minor change specific to standalone wands, otherwise looks good!

source/NeutronaWand/NeutronaWand.ino
gpstar81: If in bench test, with mode original (stand alone wand with no pack), the ion arm switch flag is always going to be on no matter what, meaning the slo-blo LED will always be lit up in that mode in bench testing, so the indicator LED does not need to be lit up.

nomakewan: Then again, thinking about it, that means the standalone wand in MODE_ORIGINAL will always have the slo-blo light on anyway, which would serve the same purpose of warning someone that there's still power going to the wand.

So I suppose it's actually fine as-is? What do you think?

gpstar81: > Then again, thinking about it, that means the standalone wand in MODE_ORIGINAL will always have the slo-blo light on anyway, which would serve the same purpose of warning someone that there's still power going to the wand.
> 
> So I suppose it's actually fine as-is? What do you think?

Yup, that was the intention.

nomakewan: Darn you github for not actually updating! Yep, okay, then this is good to go!

nomakewan (approved):

source/NeutronaWand/NeutronaWand.ino
nomakewan: What we're saying is that this is only used to indicate to the human (not necessarily the actor in the Ghostbusters universe) that they forgot to fully power-down the wand. Because in MODE_SUPER_HERO if you just turn everything off but leave the battery connected, there's no way just by looking at it that you've done so.

Whereas in MODE_ORIGINAL the SLO-BLO light will remain on, so it'll be obvious that you left it connected to power.

gpstar81: I had a request from someone who said they accidentally left the battery on over night as they forgot about it. Thus having a led blink will notify the person that the battery is still on and feeding power to the system. 

In mode original, the slo-blo comes on as soon as the ion on switch is turned on, so that acts as a safety reminder in a sense. However when the ion-on switch is in the off position in mode original, then the blinking led system will resume for it.

DustinGrau: Ah, now that makes more sense. It really is a reminder to the owner that they've left things powered up from the battery, and this adds an additional indication when there is normally none.

PR #231, opened 2023-12-30
Wand code cleanup

nomakewan:
This cleans up the fix from #230 to use a single new function called getSystemYearMode() rather than creating an entire new enum variable and copying the same conditional statement multiple times. Should improve readability, and also reduces code size.

Includes newly-compiled binaries to match.


DustinGrau (approved):
Definitely a worthwhile cleanup. Looks sane to me, so I'll merge and test with my current work.

PR #232, opened 2023-12-30
ramp down interpolation changed to circular in

gpstar81:
Ramp down interpolation changed to circular in, to match the circular out interpolation of the ramp up.


gpstar81:
After more testing and confirming, changed to quartic.

DustinGrau (approved):

PR #233, opened 2024-01-01
Fix max volume check

nomakewan:
Current check will only enforce i_volume_abs_max setting if the requested volume is beyond +10dB; this fixes the check so that it will enforce the setting if the requested volume is higher than the setting.


DustinGrau (approved):
Looks simple enough to me.

PR #234, opened 2024-01-01
Include File Refactoring

DustinGrau:
This is more of a cleanup effort than introduction of new features, but it will make the next set of changes much easier to review. Details as follows...

Background: The primary reason for this change is related to the upcoming changes to the Attenuator to support the WebUI for managing settings. That feature will add many more lines of code in the serial coms logic to perform mappings between variables and new struct objects, and the introduction of new includes provides more ease of maintenance. As a secondary reason, it allows more efficient editing when moving between files since the multi-tab approach of ArduinoIDE allows you to keep a cursor in place on one file while you navigate another.

Changes: In this branch there is a significant refactoring of source code to make the code more manageable. All logic related to the serial communications is moved to a new **Serial.h** for the pack and wand. Similarly, the EEPROM logic is moved into a **Preferences.h** following the same pattern. There are some variables and functions renamed to be more consistent between the pack and wand. There were also some duplicate function names found which used the same name but different casing and somehow ArduinoIDE allowed this to happen without warnings--this is likewise resolved.


gpstar81 (approved):
Looks good.

gpstar81 (approved):
Looks good.

PR #235, opened 2024-01-15
WebUI Preferences

DustinGrau:
This is a major enhancement effort to bring the EEPROM controls to the ESP32-based Attenuator/Wireless device, so they can be modified using a graphical interface. As part of this support was a significant logic update to the serial communications which also brought additional functionality. Some of the notable enhancements include:
- Dedicated preferences for pack, wand, and smoke features
- Controls grouped by device/purpose for ease of use
- Real-time volume updates (master/effects)
- Pack voltage readout with health estimator
- Usability improvements by way of better layout for some controls
- Consistent terminology on UI labels (better matches operation guide)

There are still some items that should be considered before merging, or as work after release:
- Testing. Testing. Testing! Beta testers may be beneficial for a small control group.
- Possibly adjusting the WebUI to be smarter about what options are enabled for use based on hardware chosen (eg. don't allow toggle for VG modes on the Power Cell if using stock Haslab LEDs).


DustinGrau:
@gpstar81 It's time to get some more eyes on this, notably yours. It's a fairly big update and I'd like any comments or suggestions now, as I've got everything seemingly working as I hoped. But it still needs plenty of testing before merging as I've touched a good bit of logic.

gpstar81:
A few bugs so far for you to track down. I will continue testing over the coming days.

Bug #1:
Web ui: Turning the Pack On not working. The pack doesn't turn on, though the attenuator thinks the pack is on (confirmed by going into the pack eeprom menu settings in the web ui)

Ion arm switch: same as above, not turning on the pack.

Pressing the Pack Off or Pack On button also sometimes but not always, makes the sloblo and orange front clippard light turn on or blink on/off.

Bug #2:
I see it is possible to set the Pack into a different year mode than the wand (Effects Theme). In the equipment status UI for the Attenuator, the Effects Theme shows what the Proton Pack is set to. The eeprom menu system, links the effects themes between the Wand and Pack to the same years when changing. 

Meanwhile it seems the EEPROM menu from the wand when changing the year mode, only affects the Pack only and not the wand.

DustinGrau:
Thank you for the sanity checks. I'll take a look at these as I get time today. I'm enabling some additional debugging in my forked code to work on these, so if you continue to test you won't be interrupted by anything I might accidentally introduce while making fixes.

nomakewan:
Apologies for poking my head in since I don't have the hardware to test with to contribute to this, but would it also be possible to test what happens with this PR if the Attenuator is removed from the equation as well?

I'm a little concerned that two people on the FB page grabbed the latest binaries from Github only to find that the Wand wasn't communicating with the Pack (with no Attenuator present), but that both the Wand BenchTest binary as well as the 3.4.0 binaries hosted on the main GPStar website resolved those issues for both users. Just want to be 100% sure nothing got accidentally overlooked in a data check somewhere or something like that.

Best of luck with the changes--they look awesome, especially for folks who troop/con with this gear!

DustinGrau:
I thought those last two people to load up the latest firmware were found to have physical wiring issues. But the wand should not be dependent on the attenuator and vice versa--the pack should be able to identify when each of the devices is connected and then send any commands as needed.

One thing I've changed in this code is the wand will initiate the handshake to the pack, and I've added code that will blink the white LED beside the vent every time it makes that attempt. So if there's a communication problem, you'll now immediately get feedback on the wand to know it has power, but can't reach the pack.

nomakewan:
This user had no wiring issues, states the fault was with the Github binary alone: https://www.facebook.com/groups/gpstartechnologies/posts/1103318744005400

I swear there was a second user who had a firmware-specific issue but I can't find their post. Perhaps they deleted it after getting the help they needed (I wish people wouldn't do that).

gpstar81:
I will give it a good test with no attenuator as well. I will dedicate another 2 hours of testing tomorrow morning without it, trying all kinds of different combinations. Still no rush as the Attenuators are in production right now and won’t arrive until the end of this month. The original shipping date was mid February, but it looks like they may be ready to start shipping out to people by the beginning of February. So still a good 2+ weeks to test and stamp out any bugs.

DustinGrau:
Just merged in some fixes along with a few other changes:
- Added some new STL's to the pack folder for housing smoke kits (Wasteland and DIY options)
- Fixed screenshot of wand preferences (removed another accidental setting duplication)
- Option to quickly turn on debugs which can help with serial debugging (off by default)

As part of my workbench testing I've tried to verify all of the items listed so far:
- Each of the devices work on their own, even without being attached.
- The wand worked as expected on its own in benchtest mode, and seeks out communication with the pack otherwise.
- The ion arm switch worked fine for me to turn the pack on or off, and the wand also worked as expected.
- The pack and wand both work as expected whether a serial1 device is present or not.
- The pack controls from the attenuator web UI worked as expected to turn the pack on or off.
- When changing the operation or year mode defaults for the pack it should propagate to the wand as expected.

Next I'm going to load everything up on my real pack/wand/attenuator and test more from there.

DustinGrau:
My final update from the evening was a cleanup of some of the API calls. There's a few that I believe were being sent before the handshakes take place, though I believe they're already handled as part of that process. Overall I think I have only the bare minimum of communications given that I've added several new ones that send more info to the attenuator/wireless. The other work was cleanup to try to make the comms logic a bit easier to maintain (hopefully).

On the UI side I adjusted some of the code to lock out the Venting button when the system is already in the overheat state, and I've confirmed (to the best of my testing so far) that all the values are flowing for the various preference screens. I also added a music volume display since I had the data and I realized that changes when the pack/wand are off and music is playing--so now you'll see that value change in realtime as well on the UI.

The one issue I've noticed is that while I can change the values for the pack's Operation Mode and the Theme Year, and despite the values being pushed via serial comms to the other devices, they don't get picked up immediately. It's only if I truly save to the EEPROM does everything get set correctly. I'm guessing there may be a few other function calls I missed which need to be triggered so the wand and pack pick up those values.

gpstar81:
Bugs:
-I am noticing some startup issues and year modes not in sync when changing settings from the webui.
Example: Have wand and pack set to toggle for year mode and operation mode at super hero. Repower the system, then change the effects them to 1984 on the pack in webui. There is then like a cross over of super hero and mode original going on. Must be a variable setting issue.

-Year mode toggle: is working for the pack but the wand is not changing over. Example, going int other wand menu system to change to 1984, while the pack year toggle switch is on afterlife and the settings are set to follow the toggle switch. The pack switches to 84 but not the wand. Might be related to the Frozen Empire enabling?


Maybe something to add:
-When Venting is disabled in the wand, have the vent button disabled in the webui.
-When no wand is connected to the pack, maybe have the attenuator wand state on the menu say: "not connected"

DustinGrau:
Good idea for the additional flag for "wand presence" as I plan to display it. I'm working on that first, along with the venting option. Then once I have those available to test I'll begin looking at the year mode issues.

DustinGrau:
Just did a quick check of the pack and wand year themes after the latest fixes, and I think I did indeed find the code I accidentally munged. The value is flowing between the devices regardless of where you set the option (EEPROM, runtime menu, web UI). The only thing I'm still battling is the runtime change for the operational mode--that seems to only work as expected when you save to the EEPROM and do a complete reboot of the devices. Again, probably just a variable not getting set correctly so I'll continue to look at that if it doesn't pop out to someone first.

gpstar81:
> Just did a quick check of the pack and wand year themes after the latest fixes, and I think I did indeed find the code I accidentally munged. The value is flowing between the devices regardless of where you set the option (EEPROM, runtime menu, web UI). The only thing I'm still battling is the runtime change for the operational mode--that seems to only work as expected when you save to the EEPROM and do a complete reboot of the devices. Again, probably just a variable not getting set correctly so I'll continue to look at that if it doesn't pop out to someone first.

1984 / 1989 year mode changes are working now from the menu. I will do some more extensive testing tomorrow morning, as well as some planned code changes.

nomakewan:
Darn. I'm kinda stumped as to what's going on with the compile check. I did update the action handler to use the actual library which contains millisDelay rather than a direct link to an outdated fork, but that hasn't resolved the issue. I want to think that maybe something changed on the action itself, but I don't see anything on the action's repository to suggest a change was pushed that matches when our action started failing, and likewise I don't see any open issues suggesting the action itself is failing en masse.

I will at least update the COMPILING_FLASHING readme to refer to the updated library though, as that will simplify things for others who might want to compile from source.

DustinGrau:
Thanks for tracking down the library issue. I had noticed that but mostly considered it a problem on the Github side with downloading the files. I've made a similar fix to the documentation for compiling the Attenuator as it uses the same libraries.

gpstar81:
I added another sound file, so update your SD cards.

Also in the process of adding all of the wand sounds to the pack playback when enabled for two reasons. 

#1 If people do not like how the small speakers in wands are overpowered by the pack, then they can enable this. 
#2 In preparation for 3D printed packs or home built packs, which lack the space for speakers in the Neutrona Wands. 

Wand bootup and shutdown sounds will now be played by the pack and in 84/89 mode as well when this is enabled. Also fixed a bug in Afterlife wand sounds playing with the pack, were never stopping when the alarm cable was going off when the lower right toggle was down.

I will try to finish adding additional sounds tomorrow. Need to get back to the audio thing I was doing yesterday for the rest of tonight.

gpstar81:
When the extra sounds feature is enabled, the Proton Pack will now play all the Neutrona Wand sounds now. This will be useful for the home builders who do not have a audio board in the Neutrona Wand, and or people who just want louder Neutrona Wand sounds overall.

Can you give it a test to see, particularly the idle beep sound in Afterlife is in sync with the Neutrona Wand for you? Also the boot error beeps as well.

I will give the Attenuator communication features some thorough testing tomorrow.

DustinGrau:
New audio changes are looking good. I just rebuilt the binaries, but won't be able to get into testing today but will try tomorrow.

nomakewan:
Pushed a minor update just now since I was in the wand code and noticed that there was incomplete code for the 48-barrel-LED option. So to put it in line with the disabled 60-LED option I've commented out the 48-LED sections as well. I also added the 29-LED mention to the Header file, and rebuilt the binaries accordingly.

DustinGrau:
Sorry about that, the 29-LED option was just for debugging an experimentation. That's NOT an officially supported value. I'll be changing some of the code back and removing the extra options for selection on the Web UI. Thanks for the reminder that those needed to come out, but nothing more needs to be done with the wand barrel stuff at this time.

nomakewan:
> @nomakewan Sorry about that, the 29-LED option was just for debugging and experimentation. That's NOT an officially supported value. I'll be changing some of the code back and removing the extra options for selection on the Web UI. Thanks for the reminder that those needed to come out, but nothing more needs to be done with the wand barrel stuff at this time.

Aha. Let me have a look at the newly-pushed changes and compare with local to see if there's anything else I've noticed. Thanks!

DustinGrau:
The code for the options will remain so I can easily enable them for myself, or try other values, though in terms of user selection they're not viable in their current form. We're looking at another option for the wand PCB and LED arrangement so the final count may change yet again by the next time time a prototype is made.

nomakewan:
Based on the above input, I've made a few additional corrections. Also found out how to compile for ESP32, which allowed me to update the documentation accordingly so others might avoid the compile errors I ran into originally.

Line 3309 in the fireStream function in NeutronaWand.ino still needs to be addressed though, as at present it will cause unexpected behavior when the 48-barrel-LED option is selected.

source/Attenuator/WandSettings.h
```diff
@@ -162,8 +162,9 @@ const char WAND_SETTINGS_page[] PROGMEM = R"=====(
       <b>LED Count:</b>
       <select id="ledWandCount" name="ledWandCount">
         <option value="0">5 - Stock</option>
-        <option value="1">29 - LED Strip</option>
+        <!--<option value="1">29 - LED Strip</option>-->
         <option value="2">48 - Frutto</option>
```
DustinGrau, line 166: The option for 48 still needs to be commented out, and the 60 is likely not going to be usable in the near future, if at all.

source/NeutronaWand/Actions.h
```diff
@@ -605,6 +605,33 @@ void checkWandAction() {
...
+                  stopEffect(S_VOICE_BARREL_LED_29);
+                  stopEffect(S_VOICE_BARREL_LED_48);
+                  stopEffect(S_VOICE_BARREL_LED_60);
+                  playEffect(S_VOICE_BARREL_LED_29);
+
+                  wandSerialSend(W_BARREL_LEDS_29);
+                  */
+                break;
+
+                case LEDS_29:
```
DustinGrau, line 629: @nomakewan The option for 29 LED should not be added as a standard option. This was ONLY for me to use to debug and experiment with off-the-shelf components. The code is still needed to operate with that count, but it should not be allowed as a viable selection to end-users.

source/NeutronaWand/NeutronaWand.ino
```diff
@@ -3310,6 +3310,9 @@ void fireStream(CRGB c_colour) {
     break;
 
     case LEDS_29:
+      // Not yet supported.
```
DustinGrau, line 3313: @nomakewan This was intentionally set. I was using the same logic for 5 LEDs to test the 29 LED setup. Please leave this as it was for internal usage.

source/ProtonPack/Serial.h
```diff
@@ -3479,28 +3479,40 @@ void handleWandCommand(uint16_t i_command, uint16_t i_value) {
...
+      //stopEffect(S_VOICE_BARREL_LED_29);
       stopEffect(S_VOICE_BARREL_LED_48);
       stopEffect(S_VOICE_BARREL_LED_60);
 
-      playEffect(S_VOICE_BARREL_LED_48);
+      playEffect(S_VOICE_BARREL_LED_5);
     break;
 
-    case W_BARREL_LEDS_60:
+    case W_BARREL_LEDS_29:
```
DustinGrau, line 3491: @nomakewan Same case here, this shouldn't be needed because this isn't a selectable option for users, and is only for internal testing.

ATTENUATOR_FLASHING.md
```diff
@@ -128,10 +128,7 @@ The following libraries are required to be installed. All but the MillisDelay li
 
 - **ArduinoJSON** by Benoit Blanchon
 - **AsyncTCP** by dvarrel
-- **ESPAsyncTCP** by dvarrel
-- **ESPAsyncWebSrv** by dvarrel
```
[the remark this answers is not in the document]
nomakewan: When I included them, it caused compile to fail completely. I then checked the compile-sketches action on this repo and found that those two libraries were not used. When I removed them from my Library Manager, the compile succeeded.

source/Attenuator/WandSettings.h
nomakewan: Would it be best to just remove all references to 60 altogether then?

ATTENUATOR_FLASHING.md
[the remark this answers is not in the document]
DustinGrau: It appears there were some updates to later libraries I used. I think I began with the "ESPAsyncWebServer" by dvarrel, but later shifted to a fork of that project that properly dealt with some issues that were lingering from the library which was untouched for years. In fact, a new version just fixed the ArduinoJson 7 compatibility issues.

source/Attenuator/WandSettings.h
DustinGrau: Just please leave 29 and 48. We don't yet know if a new version of the 48 will be coming, or if the LED count will change. I've been testing with 29 just to play around with timings and animations, but more importantly experimenting with heat dissipation methods. The biggest issue seen so far is that increasing the LEDs density increases power consumption, and that in turn becomes significant heat. I had been testing materials to see if there's something that can be safely used to mitigate that effect.

ATTENUATOR_FLASHING.md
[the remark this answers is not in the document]
DustinGrau: Confirmed, those libraries must have been either rolled into another I ended up using, or just never removed from the docs as I don't seem to see them in my list of libraries. I've fixed the docs further on my side and will merge an update for this file soon.

nomakewan:
Pushed update to address the feedback.

I've left the 60 LED sounds voice file in the sounds folder as well as in MusicSounds.h since drag-and-drop updating of existing SD cards won't remove that older file, which could lead to file total mismatches when users update their SD cards but don't explicitly delete that sound effect. Best to just leave it alone so as not to cause confusion going forward.

DustinGrau:
I carried through changes to the test compile script and documentation to match the correct "ESP Async WebServer" package name, which avoids need for a GitHub repo download each time. Also added context in docs for ESP builds, offering an alternative for the ElegantOTA package with code adjustments already made (just as used for the compilation tests). And since the optional wand barrel configuration isn't yet public, I reversed some changes to the numeric ID's used for those options as it would conflict with EEPROM values already in use.

nomakewan:
Quick question, but with the "Preferences" library by Volodymyr Shymanskyy it would appear that this add-on library is being ignored in favor of the built-in ESP32 version of the library both when compile-sketches runs and when I run a compile operation on my IDE locally. Is there a reason to use this library? And if so, is there a way to assert its use over the internal ESP32 version?

DustinGrau:
Odd, during my compile without the Preferences I got an error. It didn't seem to pick up any defaults for me, if that's the case. I'll try again, maybe starting from scratch with a fresh ArduinoIDE install.

DustinGrau:
I completely removed my libraries and re-added them 1 at a time as needed. Turns out that AltSoftSerial is no longer a requirement, and the Preferences does get picked up from the ESP32 board as stated. Docs updated, binaries and compiler script updated.

ATTENUATOR_FLASHING.md
```diff
@@ -1,151 +1,151 @@
...
-	`#define ELEGANTOTA_USE_ASYNC_WEBSERVER 1`
-1. Save the changes to the `ElegantOTA.h` file.
-
-**Alternative:**
-
-A fork of the `ElegantOTA` library can be downloaded from GitHub which contains the above modifications for use with the ESPAsyncWebServer package. Download the code as a zip use the `Sketch -> Add .ZIP Library` option to import the downloaded file.
-[https://github.com/DustinGrau/ElegantOTA.git](https://github.com/DustinGrau/ElegantOTA.git)
-
-No further configuration is needed for this library.
+- **ESP Async WebServer** by Mathieu Carbou
```
DustinGrau, line 131: What's interesting is that in ArduinoIDE it gives the name "me-no-dev" which I wanted to keep to avoid confusion. Though the "More Info" link takes you to the GitHub repo by mathieu Carbou and reflects the expected 2.5.1 version. I don't know if it's more confusing to change this or leave it as what ArduinoIDE shows.

nomakewan: That's odd; when I searched for it in the Library Manager the name came up as Mathieu Carbou. Screenshot attached.

espasync

nomakewan:
Added in the fix from https://github.com/lpaseen/ht16k33/pull/11 to the setBrightness() function in the ht16k33 library since we're using a local copy now anyway.

gpstar81:
Attenuator user interface bugs:

-Proton Pack Settings -> Changing from Super Hero to Mode Original or vice versa, makes no changes to the system when saving or doing the update. A full power down (no battery) is needed for it to apply.

-If the Proton pack was in Slime or any other mode other than Proton. Then shut the pack down and you go into the Neutrona Wand settings and change the firing mode from Video Game to CTS or CTS Mix, then you will be stuck in the video game mode you are last as well as unintended sound effects not stopping (ie: do a cross the streams while in slime mode). What needs to be done is when CTS or CTS/Mix is selected for the Default Firing Mode, then force the Proton Pack and Neutrona Wand back into Proton Mode.

-Music default not showing in the music list (This is with only the default track only, and only a pack wav trigger connected). Playback still works fine.

-While the pack is venting and you turn the pack off. The vent button never becomes clickable again even after turning the pack and wand back on.

DustinGrau:
> -Proton Pack Settings -> Changing from Super Hero to Mode Original or vice versa, makes no changes to the system when saving or doing the update. A full power down (no battery) is needed for it to apply.

I'm glad you noticed this as well. It was curious as to why the year mode could change easily but not the operation mode. I think there's something that's run in the initial setup or EEPROM read that may not be triggered later. Will investigate that, as I think I know where to look.

> -If the Proton pack was in Slime or any other mode other than Proton. Then shut the pack down and you go into the Neutrona Wand settings and change the firing mode from Video Game to CTS or CTS Mix, then you will be stuck in the video game mode you are last as well as unintended sound effects not stopping (ie: do a cross the streams while in slime mode). What needs to be done is when CTS or CTS/Mix is selected for the Default Firing Mode, then force the Proton Pack and Neutrona Wand back into Proton Mode. Note that this only happens with the Attenuator UI. In the Neutrona Wand menu system, I set it so it forces the system back to Proton to prevent this bug.

I'll set the proper defaults when I detect that the VG/CTS modes have been altered.

> -Music track number not showing in the music list. This is with only the default track only, no other music tracks on the Micro SD Card and only a pack wav trigger connected. Playback still works fine. I can confirm if you have more than 1 music track, then the track listing displays properly.

Ah, so it's when there's only a single music file present? I think that's just a math error and should be easy to fix.

> -While the pack is venting and you turn the pack off. The vent button never becomes clickable again even after turning the pack and wand back on.

If you were to use the "Reset" option to reboot the ESP device, does the UI return to normal? Either way, I think I know what's causing that, and should be able to add an action that re-checks the right variables as the pack state is modified.

gpstar81:
> > -Proton Pack Settings -> Changing from Super Hero to Mode Original or vice versa, makes no changes to the system when saving or doing the update. A full power down (no battery) is needed for it to apply.
> 
> I'm glad you noticed this as well. It was curious as to why the year mode could change easily but not the operation mode. I think there's something that's run in the initial setup or EEPROM read that may not be triggered later. Will investigate that, as I think I know where to look.
> 
> > -If the Proton pack was in Slime or any other mode other than Proton. Then shut the pack down and you go into the Neutrona Wand settings and change the firing mode from Video Game to CTS or CTS Mix, then you will be stuck in the video game mode you are last as well as unintended sound effects not stopping (ie: do a cross the streams while in slime mode). What needs to be done is when CTS or CTS/Mix is selected for the Default Firing Mode, then force the Proton Pack and Neutrona Wand back into Proton Mode. Note that this only happens with the Attenuator UI. In the Neutrona Wand menu system, I set it so it forces the system back to Proton to prevent this bug.
> 
> I'll set the proper defaults when I detect that the VG/CTS modes have been altered.
> 
> > -Music track number not showing in the music list. This is with only the default track only, no other music tracks on the Micro SD Card and only a pack wav trigger connected. Playback still works fine. I can confirm if you have more than 1 music track, then the track listing displays properly.
> 
> Ah, so it's when there's only a single music file present? I think that's just a math error and should be easy to fix.
> 
> > -While the pack is venting and you turn the pack off. The vent button never becomes clickable again even after turning the pack and wand back on.
> 
> If you were to use the "Reset" option to reboot the ESP device, does the UI return to normal? Either way, I think I know what's causing that, and should be able to add an action that re-checks the right variables as the pack state is modified.

Yup just a single music track only. Adding more than 1 and it displays the list ok.
Yes clicking on Reset/Resync fixes the vent button problem.

For mode original and super hero changes, my guess is PACKET_PACK in the pack serial.h. It looks like it is telling the wand to go into super hero when it should be mode original and vice versa.
Screenshot 2024-01-27 at 21 21 12

nomakewan:
With the music UI thing, I believe it's the check on line 198 in Index.h in the Attenuator. The current check says:

`if (musicStart > 0 && musicEnd < 1000 && musicEnd > musicStart && (musicTrackMax != musicEnd || musicTrackCurrent != musicCurrent))`

But it's perfectly possible for musicEnd == musicStart if there's only a single music track. With a single track, musicStart is 500, musicEnd is 500.

DustinGrau:
Working on those bug fixes this morning, should be able to test shortly and I'll push over to this branch.

DustinGrau:
I believe I've now fixed the recently discovered issues, and made some improvements to the UI overall (better button state handling, font fixes). The states for firing and modes should, above all, be handled properly now. And this is with the goal that you should only need to update the settings for them to become immediately effective, and not need to save to EEPROM and restart the entire system.

nomakewan:
Just a heads up that the simple ht16k33 library just updated and officially released 1.0.2, which should be picked up by the Arduino Library Manager later today. As such, no further reason to integrate that library into this repo, I think?

DustinGrau:
> Just a heads up that the simple ht16k33 library just updated and officially released 1.0.2, which should be picked up by the Arduino Library Manager later today. As such, no further reason to integrate that library into this repo, I think?

Good to know, and if it resolves the compilation issues, possibly. Otherwise, it's a very small library and we are including the source per the GPL 3 license. I'll check the latest source in a bit or give it time to propagate.

gpstar81:
Ok the previous bugs are fixed but the vent button being disabled is still there. 

Press vent then while venting, manually turn the wand off. I also notice now after doing this, while turning the wand on manually, the Pack Off button is disabled now until the wand is turned off.

DustinGrau:
> Ok the previous bugs are fixed but the vent button being disabled is still there. 
> 
> 
> 
> Press vent then while venting, manually turn the wand off. I also notice now after doing this, while turning the wand on manually, the Pack Off button is disabled now until the wand is turned off.


Ah, I don't think I caught the condition about turning off the wand, so I'll try that test to see what's going on. Good to know the others seem more or less fixed.

nomakewan:
Actually I think the bug here is in the pack itself. Tracing the code by hand, it would appear that the Attenuator waits for A_OVERHEATING_FINISHED. But the problem is that that message is only sent in a single place under a single condition, which is when the pack's overheating timer expires.

If you instead shut down the wand, it shuts down the pack, which resets the pack's internal variable b_overheating without touching cyclotronControl() which is what handles the overheat signal. Once the pack turns back on, b_overheating will be false, and so it will never go down the code path that sends A_OVERHEATING_FINISHED.

DustinGrau:
Great catch on the API call, that seems to have resolved the venting bug when the wand is turned off. I added a call to make sure the overheat state is cleared when the wand turns off and that did it.

DustinGrau:
I've just pushed a REALLY cool new feature: external wifi support! There's a new Network page (Change Wifi Settings) that will allow you to specify a wifi network to join so your pack appears as a device on that network. You can override the IP address settings, or it will otherwise pick up from DHCP like any other device.

I'll need to make some screenshots and add the documentation, but basically the device will still start with its own internal AP which you can access at 192.168.1.2 and then go to this screen to add your preferred network. The device will reboot and connect to your network. Using the IP from before, you can visit the page and find out what your new device IP is for the preferred network. With this change, you can then access your proton pack via your wifi network of choice and should solve the issue of the random disconnects due to phones wanting to change networks without warning!

nomakewan:
Pushed a documentation update since we removed the 60 barrel LED option. Also made some minor changes to some code formatting earlier, no big deal there.

Since someone pointed out that vibration settings were missing from the EEPROM menu, perhaps to make room for that (and potential future additional options) we could add an EEPROM LED Menu Level 2, and move all the remaining LED options from the EEPROM Configuration Menu to the EEPROM LED Menu? Stuff like Cyclotron LED Control, Neutrona Wand Barrel LED Count, Cyclotron 3 LED Toggle, etc?

DustinGrau:
Before we modify the EEPROM menus further, I would ideally like to see if we can get this big change released first. I know the menu management is a bit complex and would prefer that to be done first (by Michael) then I can follow up with changes to the UI.

gpstar81:
I am going to add the vibration control into the eeprom tomorrow. I am just unsure where it is going to fit in the menu at the moment. I think I can squeeze it into the existing config menu without moving anything else around.

DustinGrau:
Just did a full system check on the real pack equipment and I'm happy with the lates UI and WiFi fixes. I think this will make people really happy when they get their hands on the new features. There's nothing else I personally can think to add at this point for the current branch.

source/NeutronaWand/NeutronaWand.ino
```diff
@@ -3140,21 +3301,11 @@ void wandBarrelHeatDown() {
...
     case LEDS_48:
-      /*
-      if(ms_firing_stream_blue.justFinished()) {
-
-        ms_firing_stream_blue.start(2);
-
-        ms_fast_led.start(1);
-      }
-      */
+      // Not yet supported.
```
nomakewan, line 3305: What about this line? This can be reached if the user uses the config menu to change their LED count to 48.

gpstar81:
> Pushed a documentation update since we removed the 60 barrel LED option. Also made some minor changes to some code formatting earlier, no big deal there.
> 
> Since someone pointed out that vibration settings were missing from the EEPROM menu, perhaps to make room for that (and potential future additional options) we could add an EEPROM LED Menu Level 2, and move all the remaining LED options from the EEPROM Configuration Menu to the EEPROM LED Menu? Stuff like Cyclotron LED Control, Neutrona Wand Barrel LED Count, Cyclotron 3 LED Toggle, etc?

Aside from shifting some things to the EEPROM LED Menu, any other ideas on how to fit it into the existing EEPROM Configuration menu? (different button combinations? maybe we move the save/wipe eeprom action which takes up a menu slot to a different button combination?)

I will note, even if it's not in the configuration menu, it can save whatever current vibration setting things are set at if they are adjusted in the runtime menu. Though I prefer to have it in the configuration menu at the same time.

nomakewan:
> > Pushed a documentation update since we removed the 60 barrel LED option. Also made some minor changes to some code formatting earlier, no big deal there.
> > Since someone pointed out that vibration settings were missing from the EEPROM menu, perhaps to make room for that (and potential future additional options) we could add an EEPROM LED Menu Level 2, and move all the remaining LED options from the EEPROM Configuration Menu to the EEPROM LED Menu? Stuff like Cyclotron LED Control, Neutrona Wand Barrel LED Count, Cyclotron 3 LED Toggle, etc?
> 
> Aside from shifting some things to the EEPROM LED Menu, any other ideas on how to fit it into the existing EEPROM Configuration menu? (different button combinations? maybe we move the save/wipe eeprom action which takes up a menu slot to a different button combination?)
> 
> I will note, even if it's not in the configuration menu, it can save whatever current vibration setting things are set at if they are adjusted in the runtime menu. Though I prefer to have it in the configuration menu at the same time.

At least for now we could replace the Barrel LED count selection with it, since no one has that hardware in hand save for devs.

nomakewan:
That sounds good to me too! If there are no ill effects on the Hasbro LED board, then I see no reason why it shouldn't just be on by default. Good catch!

DustinGrau:
> That sounds good to me too! If there are no ill effects on the Hasbro LED board, then I see no reason why it shouldn't just be on by default. Good catch!

Also, I'm still seeing a lot of finished wands going the route of removing the vent diffuser entirely, so it's going to be brighter even with the stock LED. That seems like the best option with the least disruption.

DustinGrau:
Since we seem to be heading in this direction, I've just enabled the EEPROM value for the pack/wand vibration options, and worked the selections into the UI and get/set logic. Just tested the changes which appear to be saving. I borrowed the logic from the EEPROM read/save which should set the variables, though I don't currently have vibration motors connected to anything so I can't test it fully.

gpstar81:
I will give it a test in the morning. There is a minor thing that will need to be added during the wand / pack synchronisation process so wand vibration settings do not get overridden.

DustinGrau:
I've been making some changes to the Attenuator code to try and address some current and potential issues down the line. For starters, I realized that there may be a chance for IP address overlaps with some home or hotspot wifi networks, so I've moved the private network range to a Class C using 10.0.0.2 as the new default. I've updated the documentation to match this and added a note that this has changed. This now allows for Class C (192.*) networks to be used without any potential conflicts internally to the device.

I've also traced back what might be a problem encountered by some users and seems to be an issue that now arises from the change in network: users may not be assigned an IP address via the built-in DHCP server of the ESP32. To combat this, I think it's just safer to have each device which connects to the private "ProtonPack_" network be assigned an IP address manually. This will add more burden on the end-user, but they can managed how their connected devices are seen by the ESP32's internal network.

Documentation has also been updated to add steps on how to use the new "Change WiFi Settings" screen to connect to an external WiFi network. I still need to add some screenshots but wanted to get in some more testing and feedback in case I need to make more changes to the UI screens--perhaps more inline help so users don't have to refer back to the documentation to remember the necessary steps. Let me know what you think of that.

nomakewan:
I assume the DHCP issue is what's causing headaches for certain Android users, then? It would be nice if we could track down the issue with the ESP32's DHCP server instead, as asking random users to use static IPs is slightly burdensome indeed. Still, if the problem is fundamental to the ESP32 and there's nothing we can do, then a static address it'll have to be.

This is a hobby Proton Pack after all, not a piece of critical infrastructure. If using static IPs resolves connection issues for 100% of end users, that's a win.

DustinGrau:
Oh, I've been looking into the DHCP thing, but it's not easily accessible to configure. The documentation covers things at a much lower level than the WiFi library exposes which is part of the problem. I either use the helper objects and properties, or a do a lot of this by hand--not worth the time and effort. Users need to have some idea of how this works and setting up a manual IP is networking 101. Annoying, but completely doable and documented.

Also, users now have 2 choices here: use the private AP to directly connect and configure and expect things to be somewhat manual in the process, or connect the device to another WiFi network and let that handle the management of IP's.

I've tested directly with the private AP, through my home WiFi, and also my iPhone's hotspot feature. While I can connect and get an IP that way, I found that I do need to leave the Maximize Compatibility setting turned on (which ensures the 2.4GHz radio is enabled). In all cases everything works as expected so long as I follow the documented recommendations.

nomakewan:
I took a break from messing around with hardware to get back to some code. I believe I've now implemented the vibration settings for the EEPROM menu, and fixed a few minor bugs and typos along the way.

However, I must note that since I do not have a Proton Pack, this absolutely needs to be tested on real hardware to make sure everything works. In addition, this update requires two new sound files to be created: S_VOICE_PROTON_PACK_VIBRATION_DEFAULT and S_VOICE_NEUTRONA_WAND_VIBRATION_DEFAULT. At the moment the definitions are all present, but the files themselves are missing.

Great job hunting down the bug with the Wand/Pack handshake!

source/NeutronaWand/Actions.h
```diff
@@ -485,8 +485,8 @@ void checkWandAction() {
 
         // Menu Level 1: Intensify: Cycle through the modes (Video Game, Cross The Streams, Cross The Streams Mix)
         // Menu Level 1: Barrel Wing Button: Enable Spectral and Holiday modes.
-        // Menu Level 2: Intensify: Vent Light Auto Intensity.
-        // Menu Level 2: Barrel Wing Button: 5 / 48 / 60 barrel LEDs.
+        // Menu Level 2: Intensify: Enable pack vibration, enable pack vibration while firing only, disable pack vibration, reset to defaults. *Note that the pack vibration switch will toggle both pack and wand vibration on or off*
```
DustinGrau, line 488: Does this mean you must enable some kind of pack vibration to also get wand vibration?

OPERATION.md
```diff
@@ -245,7 +245,7 @@ To access the Menu Level 2, use the `Top Dial` on the top of the wand to move do
 | Option | Purpose | Intensify Button | Barrel Wing Button |
 |:------:|---------|------------------|--------------------|
 | 5 | Firing&nbsp;Mode&nbsp;Selection | Cycle through VG/CTS operation modes<sup>4</sup> | Cycle through VG colour modes <sup>5</sup> |
-| 4 | Smoke/Overheat&nbsp;Toggles | Enable/Disable Smoke Effects | Enable/Disable Overheating |
+| 4 | Overheating / Smoke | Enable/Disable Overheating Ability | Enable/Disable Smoke Effects |
```
DustinGrau, line 248: Curious as to why this was simply a swap of the options? Does it feel more logical this way?

DustinGrau:
I didn't have time to write up last night's changes so I'll provide context on that now. What I found with hot-swapping the wand was a chance that even once re-connected the handshake could get "lost" and the wand would not register with the pack. Originally when I refactored the serial data for the preferences I was trying to let the pack check for only the attenuator/wireless device and leave a connected wand to tell the pack when it was present. Turns out that some amount of bi-directional checking is in the best interest of the user. So now if the handshake fails to come from the wand the pack will have it's own backup mechanism to check. I also adjusted the debugging sounds on the pack so you'll hear different audio cues when a connected wand first synchronizes, sends a handshake, or the pack sends the handshake. Essentially it will sound like a true heartbeat to help confirm operation.

DustinGrau:
@gpstar81 can you create the 2 sound (voice) files mentioned for the latest EEPROM fix? I'm not sure what tool/service you use to generate these.
S_VOICE_PROTON_PACK_VIBRATION_DEFAULT
S_VOICE_NEUTRONA_WAND_VIBRATION_DEFAULT

source/NeutronaWand/Actions.h
nomakewan: No; it means you can have separate settings for the Pack and Wand.

DustinGrau:
New changes, old problems. So the serial logic is kicking my @$$ on the neutrona wand at the moment. I did find some areas for improvement (ie. erroneously left in a debug option) but the main problem seems to be a multitude of messages stepping on each other when they hit the pack. I'm treating any API calls from the wand as proof-of-life and therefore resetting the timers related to the handshake so the number of calls are reduced. I'm also experimenting with dropping the baud rate from 9600 to 4800 in attempt to guarantee delivery between the wand and pack. The most critical of issues I'm trying to solve is that heavy use of the wand was causing maybe 1 in 5 actions to not be registered by the pack--whether this was turning on the pack via the wand, or (more typically) just repeatedly pressing the main firing button. In my tests the pack could get stuck firing after receiving that message but missing the API call to stop firing. If things can at least hold stable for the time being, I'll be trying this out at an event this weekend and will make note of any issues encountered.

DustinGrau:
Restored baud rate back to 9600 as that does not appear to be the culprit. I did find some mis-matching between the ENUM value used to define the packet ID's, which are expected to be uint8_t (int) types. Though again that's not a smoking gun here either. And disconnecting the ESP32 does not appear to change the types of issues seen, nor does it increase or decrease severity or frequency. At best, they seem to occur at random or I've not yet noticed a pattern.

For one such issue I can enable debugs and watch the API calls for the wand and pack. When performing a 1 second firing, the wand will send the W_FIRING along with a few other codes and end with W_FIRING_STOPPED. The pack will receive every code except for the W_FIRING_STOPPED and so it get stuck in the firing state until you fire again and it hopefully picks up that previously missed code. This doesn't happen every time, but just enough to be noticed as an end-user and thus very annoying. Other codes may be missed at random as well, such as changing modes, requesting the new wand prefs, or even turning the rotary encoder on the wand. Events which flow the other direction don't appear to have issues, but there's also so few that it's hard to tell if there's a pattern. An easy one is to trigger the ribbon cable disconnect, and that is seen by the wand every time (so far as I can tell). So if anything that's pointing to an issue in the pack code with receiving commands, and only seems to be from the wand.

DustinGrau:
Oh FFS... After catching several minor inconsistencies it appears the culprit is somewhere in my newer code that I've been testing which adds the "sparking" inner cyclotron lights. I knew there was a reason that wasn't ready for public use and now I know I've got something's gone wrong in there. I was working off of that thinking there was nothing that could possibly have gone wrong, but using what's in this branch yields the expected results. At least now I know, and it confirms this codebase is still on the right track for a potential release.

Still, the latest code changes will standardize some variables names on the pack, and consistent datatypes across devices on all architectures. Commands will be sent at unit8_t (byte) to reduce the data packet size, though this means the API options top out at 254 and the wand is over 200 at the moment. As of right now I'm no longer losing the packets I had noticed previously and it even corrected some glitches I was seeing with passing the wand config data back through the chain of devices to the ESP. Everything feels snappy and responsive again compared to what I was previously experiencing.

DustinGrau:
Just loaded on my full pack (not just the bench rig), and its working great now. The latest changes include smarter handling of restarting the external WiFi network connection instead of rebooting the entire ESP device. This allows you to instantly disable the wifi, or change wifi, and get immediate feedback on the new IP address assigned as a client. For my tests I've successfully paired to my phone's hotspot and can see the internal IP address. Yesterday I was in a convention hall which was saturated with wifi networks and got errors from the phone that it could not utilize the private AP due to too much overlap with existing networks--something I have never seen before. Using the hotspot allows me to get back to the device directly and seems to be more consistent/stable which is what we really wanted/needed from this fix.

nomakewan:
That sounds awesome, and great job as always tracking down nasty serial data bugs! I had a peek at the new code and really like the refactor here; super clean. I think we finally have all the variable names matching between the various devices. Also, since switching the barrel LED was removed from the wand menus, the fact that part of the 48-LED barrel code is missing is no longer a blocking issue in my opinion, so I too feel comfortable signing off on this PR as-is.

The last thing I would ask is something Dustin mentioned earlier; do we want the EEPROM version of the vibration settings to differ from the volatile config version?

At the moment, say you set the Proton Pack to VIBRATION_NONE and the Wand to VIBRATION_FIRING. If you turn the physical vibration switch inside the Pack's cyclotron off, both devices will stop vibrating. Turn it on, and the Wand will go back to vibrating when firing while the Pack does nothing. This works the same way whether you made the switch temporarily via the configuration menu, or persistently via the EEPROM menu.

While writing up the EEPROM version I considered having the EEPROM versions of the configuration completely ignore the state of the Pack's vibration switch, but I decided against it so that if someone sets both devices to VIBRATION_ALWAYS they will still have a way to physically cut off the motors should they decide to do so while out and about without having to change the EEPROM setting. That said, definitely open to opinions for this one.

nomakewan:
> That feels like a likely case, and grounds for keeping the switch functionality in as a master kill switch for both devices. Though it begs the question of what the "toggle" option implies if the user selects that for either/both devices. Since I don't have any of the original equipment nor a vibration motor in my pack or wand, I can't recall what the normal behavior is when the vibration toggle is on.

The factory Hasbro function of that switch turns vibration on always when flipped on, and off always when flipped off. However, it only did this for the Pack as the Wand had no way to enable/disable vibration.

The default for GPStar, if I recall correctly, is "vibration while firing" for both the Pack and Wand when the switch is toggled on, and vibration off when toggled off.

gpstar81:
I actually did up some code that allows the pack to randomly let out some smoke while it is idling every X amount of minutes depending on the power level, that was requested from someone.  I haven’t got around to push over yet. The problem is there is no menu space to put a toggle for it.

nomakewan:
Might be good to keep that for the update after this one, then? I think the plan was that this one should go out, then the next update would include a menu refactor (like moving more LED options out of the main EEPROM menu and into a new level for the EEPROM LED menu). That alone would free up space in the main EEPROM menu.

nomakewan:
Just pushed a bit of a bugfix for a few things I noticed while deciding to walk through the two operation menu documentation files and compare them to the existing codebase. A few menu options were backwards from the documentation, and some of the menu options that should work in standalone (switching between Super Hero and Original, adjusting system volume) didn't. Plus some other really minor stuff.

I think that about does it.

DustinGrau:
I had asked Mario Rossi to test out the latest code since he was already beta testing a recent build. He found some odd behavior which is denying his wand from a full sync. I think the pack believes it is there and tells the attenuator as much, but the wand isn't getting the same message. I've checked in some potential changes but have yet to test them--I'll need to do it after hours but wanted to get in some updates during lunch while fresh on my mind.

DustinGrau:
So I ended up refactoring a bit of code due to the unexpected wand behavior, all to find a bit of extra copy-pasta from last night's fixes. There was an infinite loop created in the switchMode() logic which was discovered after I realized the device was going into a reboot cycle. Not a big deal overall but worth noting that we should probably get this code moved forward if it proves good, then we can work on future fixes in smaller commits.

On the plus side, I've moved to using an ENUM to set the state of the wand's synchronization state which avoids the need for multiple BOOL variables. I hope this will play nice with the benchtest mode so I would encourage re-testing of that feature.

nomakewan:
Whoops, that was my mistake. That's what I get for using find->replace and not checking to make sure that the code I'd replaced wasn't actually used in the exact same function definition that it was calling. D'oh... I apologize for wasting everyone's time with such a boneheaded move.

I was thinking of quickly implementing a way for a standalone wand to switch the "ion arm switch" on and off when in MODE_ORIGINAL with the wand status in WAND_OFF but I will keep that for a separate smaller commit after this goes out the door.

Once again, I apologize for the mess.

DustinGrau:
It's all good, as I learned a key debugging technique. Despite the numerous Serial output statements I'm now aware of how painfully slow those are to output. The root issue would cause a reboot before the final lines could be written and thus it completely threw off my sense of where the problem was occurring. Adding a brief delay() after each statement allowed them to be written and I could see I had made it past the areas I thought were the problem. Once I had ruled out the immediate stuff I found that the setup() method was being called again which clearly indicated a crash and restart, so I moved to looking for any illegal operations. That was enough to pin-point any code that looked sus and found the recursive call. On the bright side, I also learned the power of the F() macro for strings and which doesn't contribute to the dynamic memory of the program.

DustinGrau:
Mario Rossi has been running some tests for me to see how things work for him. He did find a glitch where the sync operation could cause a voice feedback on the wand. This was a goof on my part due to the changes I put in yesterday so I've hopefully fixed that. He also noted the same crackling sound with the wand's hum loop so I made some very light changes to some of those sounds (applied a high-pass filter to remove 100Hz and lower frequencies, and a light compressor setting to take the edge off the louder portions of the sounds). He also noted some good text changes for the web UI to help with usability so I've implemented those along with a shuffling of a few options on the wand preferences page.

nomakewan:
Sorry for the delay, today ended up a bit more hectic than anticipated. I was able to get back to the lab and flash the latest build onto my bench setup, and it fired right up without any issues. Switched through modes, menu, all that good jazz. There's a 'beep' like the Afterlife beep when you first fire up the wand with all three switches in 1984 mode that I don't remember if it was there before or not (been a while since I've actually fired up this setup!) but other than noticing that single beep, it seems to be awesome.

Looked at the code and it appears that in non-Afterlife/FE modes, soundBeepLoop() is called once when the wand is fired up, and the Afterlife beep plays once. Since this is explicitly laid out in the code (differentiating between the beep looping or playing only once depending on whether the wand is in 1984/1989 or Afterlife/FE) I must assume it is correct.

So yeah, cheers! Looks good on my end!

DustinGrau:
Michael mentioned the shipping dates are still holding for the hardware/kits, so assuming EOW for those to be prepared we could be on track to have this software out by the time kits arrive. If the forthcoming devices ship with the stable but old software from the current public release that should be fine, and would be preferable as I'll be out of town the next few days and won't be able to help with any software fixes until Monday of next week at the earliest.

gpstar81:
I am expecting the last bits of the first kits to arrive in 2 days and will be prepping them over the next 2 days after that before they start rolling out to people. Do you still recommend shipping with the December release firmware on those boards and let users do the OTA updates when they have it in their hands?

DustinGrau:
> Do you still recommend shipping with the December release firmware on those boards and let users do the OTA updates when they have it in their hands?

Yes, I think that would just be easier for everyone. We know there can be upgrade pains with flashing the wand, the pack, and updating sounds on the SD cards. Using the current software release that everyone should already have loaded on their equipment will ensure that the kits can be assembled, plugged in, and guaranteed to work with the existing software. Once users confirm successful operation they can then upgrade at their convenience. Sounds sane enough?

DustinGrau:
Very small change, just ensures the lower/upper LEDs are set to black at setup. With that, every device should be in a known state after configuration via setup() and binaries are rebuilt fresh to ensure they are using the latest code.

gpstar81 (approved):

PR #244, opened 2024-02-22
Bench Test Wand Ion Arm and Afterlife SFX fix

nomakewan:
This update allows a Neutrona Wand with the Bench Test firmware to change the state of the simulated Proton Pack's ion arm switch by double-tapping the Intensify switch while the Wand is turned off.


nomakewan:
Okay, all done. Wanted to make sure there wasn't any way to get into a non-deterministic state after trying this out on my bench setup.

So the way this works is if the Neutrona Wand is using the Bench Test firmware and is set to MODE_ORIGINAL, then as long as the wand is in MODE_OFF (and not in any menus) you can double-click the Intensify button to pretend that you flipped the ion arm switch on the Proton Pack. This way the SLO-BLO light isn't always on. If you click the Intensify button an odd number of times, a timeout timer will run that lasts 3 seconds, and if you don't interact with the Intensify switch in those 3 seconds it'll reset the click count to 0 and stop the timer. This way we never have a timer that runs for more than 3 seconds, and we'll never have an Intensify switch in a non-deterministic state when left alone.

The reason the non-Bench-Test firmware binary has updates is because this PR added resets for the Intensify counter to turning the wand off or exiting the menus, which isn't unique to Bench Test, and also adds that double-click timer to the Header file. Otherwise, the base functionality is unaffected since this is only for Bench Test.

gpstar81 (approved):
Works good.

Has anybody look into why in bench test mode the 16 sound effect is played when the wand gets power?

nomakewan:
> Has anybody look into why in bench test mode the 16 sound effect is played when the wand activate switch is used?

I haven't experienced this on my bench setup. Do you mean if you connect a "Bench Test" flashed wand to a GPStar Proton Pack? Or is there an edge case I haven't accounted for in my testing?

gpstar81:
> > Has anybody look into why in bench test mode the 16 sound effect is played when the wand activate switch is used?
> 
> I haven't experienced this on my bench setup. Do you mean if you connect a "Bench Test" flashed wand to a GPStar Proton Pack? Or is there an edge case I haven't accounted for in my testing?

Bench test only, no pack connected. I will send you a video.

PR #246, opened 2024-02-23
Command API Fix

DustinGrau:
This restores the start/end values for the command and message API calls to more accurately identify valid requests. This combats an issue which results from a "floating input" on the serial Tx/Rx lines which could result in phantom commands being received by the Proton pack when the Neutrona Wand is disconnected.


nomakewan (approved):
Apologies for the delay in responding, I was out in the afternoon until just now. The only qualms I had with this PR were resolved while I was out, so I'm signing off on it.

That being said, as always I have to remind you that I don't have a Proton Pack to test with, so the most important part of this PR--whether it truly resolves the phantom messages from Wand to Pack--will have to be confirmed by @gpstar81 before I can recommend merging.

Great work!

gpstar81 (approved):

PR #254, opened 2024-03-09
Attenuator Settings

DustinGrau:
This set of changes focuses primarily on bringing more user preferences to the Attenuator device itself. Users can now opt to turn off the piezo buzzer and/or the vibration motor entirely, or can disable them for specific events such as the overheat/vent sequence. Additionally, users can opt to use the vibration effect as physical feedback while throwing a stream from the Neutrona Wand.

NOTE: Continued use of the small vibration motor can cause it to become warm and repeated prolonged use could shorten its lifespan. Please enable the "Feedback when Firing" option with care.

This also includes some updates to STL files such as the latest design for the radiation indicator, and new models contributed by Dom Lake for the Cyclotron cake. Documentation and binaries have been updated for recent changes. Some variable names have been changed in the Proton Pack logic ahead of a potential change to the inner cyclotron lighting for Frozen Empire--this update makes clear that the current LED options are for the cake.


gpstar81:
Maybe should merge the audio branch into this pull request?

gpstar81 (approved):

PR #255, opened 2024-03-11
Neutrona Wand switch handler overhaul

nomakewan:
This changes the Barrel Wing Button and the Barrel Safety Switch over to the ezButton library to correct a bug reported by Ian Gendel where the Barrel Wing Button was not working for menu options which used the rotary encoder.

Additionally this removes all the double-dipping switch debounce code from the wand, since all ezButtons have built-in debouncing and as such there is no reason to run additional debounce timers on them. This greatly simplifies the code and makes conditional statements significantly easier to read.

Haven't looked at the switch code for the Attenuator and Pack yet, but the wand should be done. I had Ian test-run the code and his initial tests showed it worked as intended and didn't seem to skip randomly while changing firing modes (as I and a few others noticed in the 4.0 branch).

EDIT: Had a look at the Pack and Attenuator and they didn't have the same "double-dipping" that the wand did, so this PR should be ready to go once it passes testing.


gpstar81 (approved):
Works good on the bench test.

PR #256, opened 2024-03-11
No Ribbon Cable Alarm feature

gpstar81:
Should be added to the menu system at some point, and the Attenuator.


DustinGrau:
I think I can add the setting to the attenuator/wireless pretty easy...taking a look now

DustinGrau:
I added the logic to toggle the setting via the Web UI and also set up the preferences to store the value through a power-off. To maintain consistency with some other preferences it made sense to rename the variable to be in the affirmative--when true the pack will make use of the ribbon cable alarm; false will disable it.

DustinGrau (approved):
Reviewed and tested on my pack and working as expected

PR #257, opened 2024-03-14
Allow wand to override pack firing sounds

nomakewan:
This update will allow the WAND_YEAR_MODE setting in a wand's EEPROM to override a pack's firing sounds so that the firing sounds will be synchronized between the wand and pack. This is done by adding WAND_YEAR_MODE as an i_value component alongside the W_SYNC_NOW serial datagram so that when a wand is first connected it will transmit its setting to the pack. It is also handled in the PACKET_WAND datagram by directly accessing the wandConfig.defaultYearModeWand property so that changing the setting via the EEPROM menu in an already-connected wand or via the attenuator will also update this setting on the pack.

Additionally this updates the advanced configuration guide with the latest layout of the Neutrona Wand's configuration header file.


nomakewan:
Additionally, while looking over the serial data handshake procedure I have a question regarding the setting of a few boolean variables in the pack's code. Can I get a sanity check?

At the moment, doWandSync() in the pack's Serial.h sets b_wand_syncing to true when called, and the code comment claims it will be set to false once the wand responds affirming the sync was successful. However, instead as soon as doWandSync() reaches the end it sets b_wand_syncing to false all by itself without waiting for any communication from the wand (something normally handled by receiving W_HANDSHAKE while b_wand_connected is true). It also sets b_wand_connected to true all by itself, even though this is supposedly handled by receiving the W_SYNCHRONIZED data packet. It would appear to me that these two lines at the end of doWandSync() should be removed so that they are handled by the positive confirmation messages from the wand instead.

Also in wandDisconnectCheck() in ProtonPack.ino b_wand_syncing gets set to true even though the code comment suggests that it is "No longer attempting to force a sync w/ wand." Is the code comment incorrect or should this actually be set to false or removed entirely?

gpstar81:
So this is working on the pack side, however the wand sounds are still playing the default based on pack year.

Also this was originally intended to change the bargraph animations and not the sounds. The idea was if people wanted different bargraph animations for the different year modes: ie: 1984/89 animations for afterlife, etc.

Maybe this feature of the sound override should be another setting separate from this? Then you can get some real crazy custom combinations this way.

nomakewan:
> So this is working on the pack side, however the wand sounds are still playing the default based on pack year.
> 
> Also this was originally intended to change the bargraph animations and not the sounds. The idea was if people wanted different bargraph animations for the different year modes: ie: 1984/89 animations for afterlife, etc.
> 
> Maybe this feature of the sound override should be another setting separate from this? Then you can get some real crazy custom combinations this way.

Oh interesting; I had checked this out based on a user report and so hadn't thought to actually check the wand side of things. If the wand doesn't do this either then yeah, gonna need more code.

The nice thing is that implementing a setting is SUPER easy thanks to the standard command API. Since sending a serial data command without a value defaults the value to 0, and since '0' in this branch represents "follow the system default," the check only has to happen once on the wand side (just check this new setting boolean before deciding whether to send W_SYNC_NOW with or without a value).

I'll get this written up in a bit. Great idea!

Any thoughts on the serial handshake commentary above?

gpstar81:
> > So this is working on the pack side, however the wand sounds are still playing the default based on pack year.
> > Also this was originally intended to change the bargraph animations and not the sounds. The idea was if people wanted different bargraph animations for the different year modes: ie: 1984/89 animations for afterlife, etc.
> > Maybe this feature of the sound override should be another setting separate from this? Then you can get some real crazy custom combinations this way.
> 
> Oh interesting; I had checked this out based on a user report and so hadn't thought to actually check the wand side of things. If the wand doesn't do this either then yeah, gonna need more code.
> 
> The nice thing is that implementing a setting is SUPER easy thanks to the standard command API. Since sending a serial data command without a value defaults the value to 0, and since '0' in this branch represents "follow the system default," the check only has to happen once on the wand side (just check this new setting boolean before deciding whether to send W_SYNC_NOW with or without a value).
> 
> I'll get this written up in a bit. Great idea!
> 
> Any thoughts on the serial handshake commentary above?

I will have to look into the serial handshake at some point soon to see. I know lots of it was refactored by Dustin with the Attenuator data sync.

DustinGrau:
I can't believe I completely missed the fact that the mix-match was for the bargraph only. In my head I thought it was also applicable to the sounds, but now I realize that if we offer that it makes the year mode even more confusing. You'll have the pack (sounds, lights), the wand bargraph, and the wand sounds all with different themes. Adding to that is the "extra wand sounds to pack" option and you've got the pack now playing mixed effects. I think while the intention is good we should put this feature on a backlog and first solve how to get more menu items into the EEPROM system. Right now we've run out of space and I have a few items living solely in the wifi web UI settings.

And speaking of those settings, something I had been considering on the serial comms is reusing the STRUCT for the settings to help with the initial device sync. Since we have a singular object with essentially a copy of the values from various boolean and integer fields for options, I've wondered if we could retrofit some of the loose variables to instead use the object.property notation and just keep to the single construct as the source of truth for settings. I've not attempted any of that given how much it affects, and not sure if it's even a viable idea long-term, though it could simplify quite a bit.

nomakewan:
The thing is, from a technical standpoint the wand is what's actually doing the firing. So at least as far as specifically the firing sound goes, I'm on board with having those sounds follow the wand's year mode selection. I'm still going to implement the boolean configuration item for it so it can be disabled if desired, and that means down the line once the menu gets reshuffled (as I've mentioned before, there are a lot of LED options in the non-LED EEPROM menu that need to be moved into a second LED EEPROM menu level) it'll be ezpz to implement there.

Since it's pretty simple to do, I'm still gonna write it up here. Even if it only ends up being reference for later, that's fine by me.

Dustin, would you mind checking the serial data handshake bit from my first comment? I just want to be sure that how it works in my head is how it works in reality (something I can't check myself for obvious reasons).

gpstar81:
One of the things that was mentioned previously was to move all light related options from the config EEPROM to a second level of the lighting EEPROM. This would free up space in the config EEPROM menu.

Some of the things still that need to go into the config EEPROM are:

-Vent Light Intensity (this was removed recently, it should be added back in so it can be disabled)
-Enable/disable the ribbon cable alarm
-Enable a random smoke timer, so the smoke pins are randomly activated while the pack is idling (based on x amount of minutes of idling). (requested by somebody)
-Independent minimum volume control (able to adjust the lowest the volume can go for either the pack or wand. This adjusts the % based offset that is currently implemented).

DustinGrau:
I know I didn't make the logic any easier for the wand sync, but I wouldn't touch any of that nor try to "correct" it right now. It's working, and that was a feat in of itself to do. As I know you don't have a pack available that stuff was very difficult to debug and nigh impossible without the pack. I'm also not feeling warm and fuzzy with sending an additional value with the W_SYNC_NOW command as that's really needed to kick that process off and isn't part of the sync of actual values. My proposal was to send the WAND_YEAR_MODE with something like W_PROTON_MODE, etc. when the stream mode changes. But if it only needs to be sent once at the sync then it could be added as part of the sync process just after W_SYNCHRONIZED is called because it's flowing from the wand back to the pack (most everything else is pack to wand).

nomakewan:
It needs to be sent three potential times.

1. The moment the wand is connected to a pack for the first time. Within 1ms of the wand powering up it will blast W_SYNC_NOW so that's why I put it here. The datagram includes a value whether you define one or not (either it will be 0 or nonzero) so it's not like there's additional overhead by sending a nonzero uint16_t.

2. When the wand saves and exits the EEPROM menu. This is in case the user just changed the setting on an already-connected wand via the EEPROM menu.

3. When the Attenuator forces an EEPROM update. This is in case the user just changed the setting on an already-connected wand via the Attenuator's EEPROM menu.

These are already implemented packside, and halfway wandside. Packside is confirmed to be working which is great. Wandside I'm almost done fixing up and will be pushing shortly so you can have a look to see what I mean.

nomakewan:
Okay, just made the changes. There is now a boolean configuration option on the wand that will enable or disable this functionality.

The wand's firing routines will also both obey this boolean and obey the appropriate year setting.

If the boolean is set to false, W_SYNC_NOW will be sent without the optional variable, which means it gets sent with '0', which the pack will correctly interpret as YEAR_DEFAULT which will automatically have it follow SYSTEM_YEAR without any further changes.

nomakewan:
As this PR is now out of sync, I am going to close it. The closed PR will remain as a point of reference for future discussion regarding the implementation fo this feature, perhaps when the menu refactor happens.

PR #259, opened 2024-03-17
Lighting Updates

DustinGrau:
This is a HUGE update in terms of milestones. This adds the official code to support the 49-LED wand barrel assembly from Frutto Technology and implements the initial animation sequences. This is fully expected to change and be enhanced as we discover new ways to utilize the additional LEDs. Though initially it gets the selections added into the wireless/attenuator web UI and makes it possible to enable with some expected default behavior.

There is a similar update to the pack to extend the LED chain for the inner cyclotron (cake) lights to wrap externally around the cake within the "cavity" for more effects. Initially this is something of a sparking effect as inspired by a pack seen in the Frozen Empire trailer. As the movie premiere has not yet happened it's hard to tell what this behavior is exactly, so the implementation may be a bit of an estimation if anything. The web UI is updated to allow selection of a value between 0 and 30 as that is the maximum that can be supported in addition to a 35-LED ring light inside of the cake.

Minor updates to the WiFi to hopefully improve stability. YMMV. For those who care to compile and build their own firmware there are now options to compile in defaults for a preferred wireless network, making the software automatically configured without need to access the private network first.


DustinGrau:
Still doing some work to wrap this up, but getting a draft PR ready with notes.

gpstar81:
What do you think of merging in some of the audio fork into this in preparation for the new audio board?

DustinGrau:
I'm hesitant on this update since I'm modding the LED counts on the wand and pack, and had already run into serial glitches with the too-high count in the pack. Both seem to work well with their current values but if something goes wrong I'll know it was my changes and not the audio updates.

DustinGrau:
One more change to the WiFi "fixes" is reducing the CPU from 240MHz to 160MHz to conserve power (it's already a dual-core chip that's faster than the Arduinos), and I've disable the WiFi sleep mode (which can use more power). I'm hoping the net consumption is the same while possibly making the networking more snappy.

DustinGrau:
Just made some adjustments to the timing on the animations for the new wand barrel. It's definitely "good enough" for the moment. Also fixed the total count logic for the sparking lights so they kick in at the right point in the chain of LEDs and for the right number of elements.

nomakewan:
Hey there! So since I don't have the prototype hardware in hand, I'm going by a video you posted to the FB group earlier. There was one glitch I noticed in the initial preview which may have already been fixed, may not, so I figure I'll ask. Since the code for the 49-LED Frutto board is essentially using a lot of the 5-LED code, this might actually also be in the Hasbro too but there are so few LEDs I'd hard to notice.

In the video without the Frozen Empire barrel shroud, the proton and slime firing streams seem to work fine. However, when throwing the stasis stream ends, instead of the colors fading from the base of the barrel to the tip, they appear to fade from the middle of the barrel to the tip. Looking at this several times, this appears to be an optical illusion caused by the "bright" part of the barrel animation effect being generated halfway up the barrel and lingering there for a moment after throwing the stream has already ended, skipping the space from the base of the barrel to the tip.

This behavior is exemplified by the meson collider, since it has no background animation and is essentially _only_ the "bright" parts. When firing the boson darts ends the dart appears to get 'stuck' halfway up the barrel rather than making it all the way to the end. When this animation ends, considering how rapid-fire it is, any bright yellow bit currently animating should always be allowed to reach the barrel tip before disappearing.

If these animation glitches have already been fixed then I apologize; I understand that there have been quite a few commits since those videos were shown off. But I would be remiss if I did not mention them here.

Amazing work, and can't wait to get my hands on one of these!

DustinGrau:
You're right, the animations are nearly identical because I copied everything then tweaked it to work with the higher pixel density. I don't expect these will be perfect or even "corrected" at this point. Since the hardware isn't up for order just yet my focus is on getting the enablement feature out, and making sure it doesn't break for the stock wand lights. Fixing the sequence can always be done later, and with hardware more available to everyone.

DustinGrau:
Some promising news from the newest code for the attenuator: it seems to have at least resolved one issue affecting a user. Given that my initial attempt didn't make a difference, that means some of the most recent changes may have been the true fix.

gpstar81 (approved):

PR #261, opened 2024-03-18
Updated sounds & fixes

nomakewan:
This PR fixes missing menu sounds for switching between Super Hero and Mode Original when the wand is in bench test, fixes incorrect fake ion arm switch setting when exiting the EEPROM menu when the wand is in bench test, adds in the ability for a bench test wand to save/load EEPROM settings for system mode and system volume, fixes overlapping shutdown sound effects when forcing a vent, fixes lag on the barrel retraction switch, adds a new GB1 wand activation sound, adds a new GB2 startup sound for bench test wands, removes a legacy configuration variable from the pack, and corrects several typographical errors.

The new GB1 sound is the one heard in the Sedgewick when Spengler exits the elevator, as well as when Ray heats up his wand to confront Slimer. It will play only in Super Hero mode when the Activate switch is flipped (matching Spengler's appearance from the elevator), and also just so happens to match the bargraph animation already present. The normal warmup will play if the vent switch is turned on afterwards. If wand boot errors are disabled and the vent switch is turned on before the Activate switch, this alternate sound will play (matching Ray's heatup versus Slimer).


gpstar81:
Can you test if music playback is broken or something changed? Before it was changed where it did not matter if the wand had the music or not, as the pack controlled track selection and playback queues. I can not get past the first track anymore, no matter if I try via the pack or wand.

nomakewan:
> Can you test if music playback is broken or something changed? Before it was changed where it did not matter if the wand had the music or not, as the pack controlled track selection and playback queues. I can not get past the first track anymore, no matter if I try via the pack or wand.

Yep, just figured it out. Most of my time was actually spent troubleshooting my setup only to find out the WAV Trigger had been told to start muted. After fixing that finding the issue was simple (track reporting had the wrong boolean flag).

All fixed in this PR.

gpstar81:
The fixed worked. All good. Though I believe reporting needs to be turned on for the WT wand side as well, for when the wand is in benchtest mode and needs to do a music track check.

nomakewan:
> Actually music start/stop is broken and the loop feature as well. You can not stop the music. I will merge this and go through the code and open a new branch and fix the issues.

Just a heads up that reporting disabled on the wand side is how at least V4.0.0 through V4.4.0 did it:
https://github.com/gpstar81/GPStar-proton-pack/blob/main/source/NeutronaWand/NeutronaWand.ino#L7833
https://github.com/gpstar81/GPStar-proton-pack/blob/V4.0.0/source/NeutronaWand/NeutronaWand.ino#L7613

And on my standalone wand I can still start, stop, and skip next/previous just fine and always could.

I suspect the problem then must be on the pack side?

Thank you for the merge!

gpstar81:
The audio branch has everything fixed now.

The reporting is used to trigger the custom firmware on the wav trigger to send back the track status of a particular track, as things can be missed during the interrupts being disabled on the pack side during updates of the lights. As long as the wand is using the same WT serial comms library, it should be ok and can be on. Since the wand has it's own track checking during music playback in benchtest mode, it should be on, or on at least for benchtest.

PR #269, opened 2024-03-25
Refactor serial handshake

nomakewan:
This PR completely refactors the serial handshake process between the Proton Pack and Neutrona Wand. It makes the actual handshake process more state-driven, increases the number of pings the Pack will wait for before starting a timeout from 1 to 2, and creates a new serial data struct specifically for synchronizing initial settings between the Pack and Wand.

Please put this through its paces and make sure nothing got fundamentally broken in the shuffle. Please note that further code efficiency is possible with this new struct in place as several Pack commands may no longer be necessary, and that this struct could also be expanded if necessary to allow other devices to share sync data as well. For example, leaning out the Pack-to-Attenuator sync process in a similar way.

### DIFF
--- a/source/Attenuator/Serial.h
+++ b/source/Attenuator/Serial.h
@@ -38,7 +38,8 @@ enum PACKET_TYPE : uint8_t {
   PACKET_DATA = 2,
   PACKET_PACK = 3,
   PACKET_WAND = 4,
-  PACKET_SMOKE = 5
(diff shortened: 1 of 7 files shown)


gpstar81 (approved):
Appears to sync up with no issues on my side during a quick test. Will give it some good testing over the days during the continued audio development.

PR #270, opened 2024-03-26
Music delay change

nomakewan:
This PR reduces the delay before checking the track status after resuming a paused track from 20 seconds to 8 seconds (to match the normal while-playing timer length), and changes the delay before playing the next track in sequence from 2 seconds to 0.5 seconds (as realistically this will already have a delay of between 0 and 8 seconds depending on how long remains in the track status check and the music system status check timers). Also clears up some whitespace issues and removes some commented-out legacy code.


gpstar81:
ah I am in the middle of fixing up some serial code, I will look at manually adding this in. Is it really up to 20seconds during a paused track?

nomakewan:
> ah I am in the middle of fixing up some serial code, I will look at manually adding this in. Is it really up to 20seconds during a paused track?

Yes. When you first start playing a music track the timer starts at 20 seconds, which is fine since just about any music is going to be longer than 20 seconds. But if you resume a paused track then at the moment it restarts the timer with 20 seconds. So say you pause a song right near the end then resume, it'll wait 20 seconds before checking the status again. Then wait another 2 seconds to play the next song.

Feel free to add it in manually! I'm open to any and all changes.

gpstar81:
Ok I will push over my changes now and let you do some things as my day is ending here.

One thing I noticed with the updated serial changes you recently did, it is no longer possible to hot swap wands as the wand stays in a loop sending command #3 which i believe is W_SYNC_NOW? 

Also the music loop command never gets sent to the pack when the wand tries to set it:
W_MUSIC_TRACK_LOOP_TOGGLE

gpstar81 (approved):

PR #271, opened 2024-04-03
Migrate to Switch

nomakewan:
This PR migrates the Pack and Wand from ezButton to Switch. This migration allows for proper deglitching and debouncing of the switches, which makes them significantly more reliable. It also includes built-in functions to handle switched states (rather than using "pushed or released"), long presses, double-clicks, and the like.

Due to the intricate nature of the interaction with the rotary switch on the Attenuator I have not attempted to migrate it over, so it is still using ezButton. Since Switch includes long-press and double-click functions built in it should be possible to migrate, but the new "lockout" functionality may require massaging. Or it can be left as ezButton too.

If the Attenuator moves to Switch, the compile-test.yml file will have to be updated accordingly.

### DIFF
--- a/.codespellignore
+++ b/.codespellignore
@@ -4,4 +4,5 @@ gpstar
 GPStar
 DOUT
 dout
+doubleclick
 doubleClick
--- a/.codespellignore
+++ b/.codespellignore
@@ -4,3 +4,4 @@ gpstar
 GPStar
 DOUT
 dout
+doubleClick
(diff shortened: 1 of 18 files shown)


nomakewan:
After discussion, have reverted the Proton Pack code to use ezButton again as the Pack has pretty hefty memory usage and Switch uses more memory than ezButton does. Further the Pack's few controls are far less timing-critical and so shouldn't cause obvious issues like the controls on the Wand do.

The Attenuator should have enough memory available to run Switch should that be desired, but for now this works.

Issue #276, opened 2024-04-11
Overheat problems


Toy203:
Just updated to the latest revision and cleaned EEPROM. Overheat function is enabled in the eeprom. 

If i fire the wand on power level one (with all levers up), overheat sequence is triggered immediately . The wand shoots for about 1 second then the pack enter overheat: alarm sounds trigger for way too short time (it only does a single beep, i believe is "034_pack_beeping.wav" or "060_beep_8.wav"), lights slow down, strobo in the n filter starts and smoke hiss is heard.

I cannot seem to be able to trigger overheat while firing with the other power levels. I used a c clamp on the intensify just to be sure, overheat doesn't seems to trigger on continuos firing. 

If i rapidly press intensify thought, it works. On level 5 if i press it 6 times overheat is triggered, but it's a weird quick overheat without the ramp down etc. Just an alarm sounds is played for a couple of seconds (it sounds like "067_beeps_low.wav") and no smoke hiss is heard or lights in the n filter.

gpstar81:
I had same issue. A power reset was required after wiping the config-EEPROM then re-configuring it again if required.

Toy203:
Power cycle seems to have fixed most of the issues. 

I just noticed the bargraph animation defaults to the regular animation before shutting down due to overheat, which is a bit clunky coming from the super hero firing animation. Lord Nitpick is back in town.

https://github.com/gpstar81/GPStar-proton-pack/assets/153134073/513cdf2e-adbd-4939-aed0-b88ec9f0fe1f

nomakewan:
> Power cycle seems to have fixed most of the issues.
> 
> I just noticed the bargraph animation defaults to the regular animation before shutting down due to overheat, which is a bit clunky coming from the super hero firing animation. Lord Nitpick is back in town.
> 20240411_120253.mp4
> 
> Also, i still can trigger that weird alarm status if i rapidly press Intensify while firing. I can find a certain over-sensibility on the intensify button too, firing state switch off easily if i wiggle the thumb a bit while pressing it. Might be my mighty soldering but didn't have this feeling before.

The bargraph animation is actually from the video game for overheat, and has been this way since the original firmware release that implemented overheating. As such it's actually intended behavior as of now. Do you have a suggestion for a change to the overheat animation in Super Hero mode?

The alarm state is intentional and was implemented a while back to prevent people from overloading the serial bus by making rapid firing changes. This is the "anti button mash" system.

[...]

PR #278, opened 2024-04-11
EEPROM, Cyclotron lid, bug fixes

nomakewan:
This PR updates the EEPROM CRC32 functionality so that its lookup table is stored in program space instead of RAM to save RAM space, and changes the location of where the EEPROM objects are stored and which memory space the CRC32 routine actually hashes. By doing this we can assure that any firmware update which modifies an EEPROM object will always result in a CRC32 mismatch and thus automatically restore the EEPROM to default values.

It also restores the 1/3 LED selection for the 1984 cyclotron lid in the EEPROM menu, as the amplification option is no longer necessary. There are also additional checks added to removing and replacing the cyclotron lid to address momentary lighting glitches when the pack is off.

Additionally this fixes an audio bug when changing smoke settings via the menu system.


nomakewan:
Latest push has the following additional changes:

-Corrected typo for datatype of cyclotron LED on status variable
-Created new PROGMEM_READ macro for reading from PROGMEM
-Fixed missing 20-LED ramp-up timer definition
-Fixed missing 36-LED voice command when toggling cyclotron LED counts
-Fixed Overload Pulse timing in Power Level 4
-Moved cyclotron LED arrays into PROGMEM (saves 160 bytes)
-Moved smoke arrays into PROGMEM (saves 40 bytes)
-Unified "power level" to refer to the wand power level throughout code

The majority of this won't affect the binaries much (they're just changes to how the code reads for humans), but the new PROGMEM changes are pretty major for the Proton Pack. By shifting the large array constants over to flash we are saving 200 bytes of RAM, which combined with the previous CRC32 change has netted a RAM savings on the Proton Pack of 268 bytes. While a call to flash takes 1 additional cycle (~62.5 ns) the array calls for CRC32 and smoke are one-offs and are unaffected. The cyclotron array calls are more numerous, but traversing the entire array is still only a time difference of 0.0025 ms, well within the fastLED window.

Please do confirm that there are no ill effects in cyclotron LED functionality, but my local calculations show that flash should be more than fast enough for these particular array calls.

nomakewan:
Okay, latest push is a slew of bugfixes.

-Adds new unique SFX for button mashing lockout
-Cleans up wand fastLED timer calls
-Fixes functionality of wand boot error mode and button mash lockout mode
-Fixes hat light 2 getting stuck on/off if user stops firing while overheat alert is playing but before overheating triggers
-Fixes idle wand sounds not recovering after overheating finishes in Afterlife/Frozen Empire modes
-Fixes overheat alert sound staying on when user stops firing before overheating triggers
-Fixes quick vent smoke not firing if overheat sync to fan is disabled (resolves #276 )
-Fixes various timer reset calls for enhanced precision
-Makes sure power alert indicator only comes on when wand is fully off and idle

These are all the bugs I've found so far. I think there are still some places where, for example, it may be advantageous to add more serial wand sound calls (that is, places where the wand alone is set to play sound but a wand that relies on a Proton Pack will play no sound), but other than that, it seems quite solid on my bench.

nomakewan:
For documentation's sake, here's what the latest commits did:

-Removed firing tail sound when button mashing lockout triggers as no firing could have occurred at this point
-Changed wand barrel addressable LED check to run in an infinite loop the same way the attenuator and pack do
-Sped up the wand tip heatup animation to better match the sound effects
-Adds new default white LED blink rate for Afterlife mode to better match the props, and adds GB1/GB2 blink rates if the wand is set to GB1 or GB2 mode (which blink rate is set when power is first applied to the wand and is retained until next power cycle)
-Corrects a few RNG calls to cover the intended range
-Corrects a few bargraph animation related calls to cover the intended range

gpstar81:
Cyclotron is not lighting up during normal operation.

When entering the eeprom LED menu, only the top 2 LEDS from the powercell light up when selecting the 15 led option. The cyclotron lights up correctly in the LED eeprom menu, but does not function in normal operation.

nomakewan:
Reverted the changes to the power level calculation with the Hasbro 5-LED bargraph. I still have some questions relating to how the bargraph calculations work, specifically for Mode Original, but they can be left for another time.

The primary change relating to cyclotron functionality was moving the matricies to PROGMEM, so of course my concern is that something nasty has happened there. However, considering this PR also made a major change to the basic functionality of the EEPROM menus I do need to be sure that the appropriate LED count setting changes were made after the first time powering up with this PR's code loaded. The new EEPROM CRC function will detect changes to the size of the EEPROM Config and EEPROM LED objects and clear the EEPROM if the CRC does not match. Additionally the location in EEPROM where these objects are stored has changed, guaranteeing that the first power-up after loading this code will result in wiping the EEPROM clear and starting from scratch.

PR #290, opened 2024-04-29
Handshake update

nomakewan:
Allows the wand to time out after 9.75 seconds of not receiving any responses from a pack, and go back to attempting a new synchronization process.


DustinGrau:
If you could please recompile the binaries (which will be needed anyway) I can test this later this evening.

nomakewan:
Closing this as premature as it introduced several bugs into the normal handshake routine which were not previously present.

PR #291, opened 2024-04-30
Audio update

nomakewan:
Changes startup master gain setting for wand back to -70, change check in handshake routine so that a WAV Trigger cannot be falsely detected as a GPStar Audio board.


nomakewan:
As a note, we have had a user with a WAV Trigger board that did not respond with its version string (RSP_VERSION_STRING) when asked. It did, however, respond with RSP_SYSTEM_INFO when asked. Because of this, the GPStar wand board thought it was talking to a GPStar Audio board instead of a WAV Trigger.

While this PR makes it so that a WAV Trigger cannot be confused for a GPStar Audio board, it does not make it so that a WAV Trigger which does not respond with RSP_VERSION_STRING will be detected. If this is desireable it would be very easy to implement, but I do not know if this is desireable and so have not implemented it.

gpstar81 (approved):

PR #293, opened 2024-05-01
Code cleanup

nomakewan:
Cleans up some code by correcting spacing, data types, variable names, and moving function calls. Also adds voice confirmation for the physical year switch since the year change does not occur immediately and only takes effect once the pack has fully shut down.


nomakewan:
Fix identified by @gpstar81 implemented with the latest commit to this PR, which resolves #294 .

gpstar81 (approved):

PR #300, opened 2024-05-01
Poweron sound

gpstar81:


gpstar81:
Power on animation to the power cell.

PR #307, opened 2024-05-05
Fixes for WiFi stability

DustinGrau:
This changeset attempts to resolve some lingering issues which have resulted in sub-par wifi connectivity. The first is a simple adjustment to reduce the number of retries to establish a connection to an external wifi network. When out of range of a network, such as at a convention, the ESP32 device would try for an extended period of time to connect to a non-existent network. Normally, the connection should take less than a couple of seconds but this could take upwards of 20 seconds at times.

The other issue has to do with the case of bad credentials or a non-available network, and allowing the device to fully transition back to the built-in AP mode only. Again, something that would be mostly seen when outside of a home or known-stable environment, and was causing the wifi to become completely unavailable on the device. This change ensures that the private network is always available, even if the environment is less than suitable (read: heavy RF interference or congestion).


DustinGrau:
Note that all binaries were rebuilt, but this should only affect the ESP32 controller in use for an attenuator or wireless adapter.

DustinGrau:
I've been running the equipment for hours on end today with no glitches. This was partly to see if my fixes also took care of an issue I noticed yesterday at an event where the attenuator became unresponsive after a few hours. Looks like I may have nixed that issue as well so I'm opening this for review and merging.

source/Attenuator/Wireless.h
```diff
@@ -332,19 +346,25 @@ bool startWiFi() {
...
+
+    // When external WiFi is unavailable, switch to only SoftAP mode.
+    WiFi.mode(WIFI_MODE_AP);
     delay(250);
   }
 
   // Start the built-in access point (softAP) with the preferred credentials.
+  // This should ALWAYS be available for direct connections to the device.
   if(!b_ap_started) {
     b_ap_started = startAccesPoint();
```
nomakewan, line 365: Just a minor typo; function name should be `startAccessPoint()`.

nomakewan (approved):
Just a little typo.

Otherwise the code looks really solid. Great to hear that it was able to run for hours on end without issue!

gpstar81 (approved):

source/Attenuator/Wireless.h
DustinGrau: Damn, never noticed that. At least it's consistently misspelled for compilation :)

PR #312, opened 2024-05-16
Web UI Enhancements

DustinGrau:
The primary changes in this set is the addition of a new music volume increase/decrease which means an update to all 3 devices to support the new serial API. This now brings all of the volume controls to the Attenuator/Wireless Web UI and puts all audio methods in the new Audio.h files. And as a matter of terminology I'm trying to get away from the public label of "Master" and move to "Overall" or "System" volume as a matter of good practice among software devs, so the documentation got a small update to match the new controls.

Speaking of controls, this update redesigns the 3 volume controls to be more consistent so they appear in the same space within the UI. The music navigation also got a more symbolic update so we can rely less on text since this is an international kit. The power level display in the system info panel now shows a graphical indication of the power and changes color to match the stream mode in use. And lastly, the battery indicator was changed to report as the Powercell status with a "GeV" label. To quote Ray Stantz, "Nothing hurts like a billion electron-volts!"

I've also standardized many of the new colour values across devices and implemented a timestamp that gets built into each compilation of the Attenuator code when using the built-in compile.sh script. This allows a quick check of what should be the latest release installed for a GPStar kit as we typically require major updates across all devices at the same time.


DustinGrau:
@nomakewan Requesting your review first as I noticed something odd after I integrated your latest fixes. When throwing a stream the system volume gets reduced to 0%. I'm thinking it might be related to the logic which reduces and mixes the audio effects when switching streams. I've gone through my new code and I'm not really touching the main volume, only the music volume logic. Any chance you can revisit some of the code you changed last since this was working yesterday?

nomakewan:
> @nomakewan Requesting your review first as I noticed something odd after I integrated your latest fixes. When throwing a stream the system volume gets reduced to 0%. I'm thinking it might be related to the logic which reduces and mixes the audio effects when switching streams. I've gone through my new code and I'm not really touching the main volume, only the music volume logic. Any chance you can revisit some of the code you changed last since this was working yesterday?

Is it happening only in Afterlife/FE, or in all years? The change I implemented regarding volume mixing was just for going between slime/stasis/proton/meson/etc and was only related to adjusting the gain of the idle pack loop in Afterlife/FE. Nothing else is touched, and I have otherwise made no volume or gain-related changes.

I did test 27dabfb on my standalone wand and did not experience any strange volume changes, so at least the wand-side code does not appear to be commanding anything. I'll go over the pack stuff to see if anything jumps out at me.

DustinGrau:
I had only tested with FE theme, as I had yesterday. There was nothing else I had changed since then for the pack/wand beyond syncing the new warm_white color. Up until the branching today I had been merging in your changes to my fork and testing with all my extra mods.

nomakewan:
Did find and fix a missing sync packet enum on the Attenuator. I don't think that's related to the volume issue, but it could still have potentially caused issues. Will continue digging.

DustinGrau:
Good catch though on the serial data. That may have caused a glitch I noticed once recently but couldn't replicate.

gpstar81:
No problems with the volume on mine with the latest main branch merge, I am running GPStar Audio's. Does it do it when the Attenuator is disconnected?

DustinGrau:
Shit. It's only happening with the Attenuator attached. F*CK, it was not doing this yesterday...

DustinGrau:
Ok, plugged in a USB cable and turned on all debugging. So far I'm not seeing any serial commands that would turn down the volume...but I also don't see the behavior since going full serial output. Let me revert and see what happens.

DustinGrau:
Welp, looks like it was either a bad flash or something transient. Both worry me for OTA use, but it's all working correctly now.

nomakewan (approved):
Looks good to me! I did check over the code and didn't see any obvious reasons that could have caused the earlier bug, so since that appears to have been resolved, I'd say it's good!

DustinGrau:
Cool, I'll merge this a bit later this morning and create a 5.1 release with notes from your extensive and fantastic mods for the VG modes, and some footnotes on these updates.

PR #324, opened 2024-06-01
Vibration fixes

nomakewan:
This PR fixes situations where a custom pack without a vibration toggle switch would turn off all vibration on both pack and wand even if the EEPROM mode is set to override, fixes a situation where a wand would disable vibration based on initial pack setting even if set to override, fixes incorrect inner cyclotron switch panel LED behavior regarding vibration on/off setting, simplifies several vibration-related checks on the wand, and improves vibration-related code documentation.


gpstar81 (approved):
Didn't test it, but looks ok.

nomakewan:
I have reached out to the two users who reported this bug to have them test it, since they know what steps they took to reproduce the issue.

Toy203:
Might be able to test it tomorrow morning (Italy time) if needed before pull.

nomakewan:
> Might be able to test it tomorrow morning (Italy time) if needed before pull.

That would be fantastic if you could!

DustinGrau (commented):
Overall I've been a bit confused by this logic, and given the effort to clean it up it's not really become that much clearer. There's a variable for "on" and "enabled" but it's confusing as to which means the feature was enabled (by the user) vs a toggle which controls override/use. Could the toggle switch state be tracked by a more obvious var name? And for the user preference, is this another place we could distill the choice down to an ENUM as a single state for operation rather than grouping logical values like a giant truth table?

nomakewan:
I have updated the variable names to make it clearly what they do.

As for making it an enum, yes, that can be done. The issue with making it an enum is that it would then take the setting out of Configuration.h and put it into Header.h and have the initial setting be done in setup() of the INO instead. With the bools the way they are, they can be set in Configuration.h instead.

While I'm okay with doing that, I'd like to get input from @gpstar81 to see if that would be okay before making that change since it means removing an Advanced Configuration option.

DustinGrau:
Thank you, that does help with the renaming. And I see that the VIBRATION_MODE_EEPROM at least condenses things down to a single setting for storage, so you're right the only thing stopping a true consolidation is the Configuration.h file. Hopefully that may go away entirely as we get more of the fine-tuning for preferences moved into the menu system. So perhaps that's more of a consideration for cleanup when that particular re-org effort takes place.

Toy203:
Let me know if you need some beta testing for the new code. =)

Toy203:
Confirmed, vibration settings are now read correctly from what i can see.

nomakewan:
The person who reported the bug the first time says they will be able to test it tomorrow. I would still like to hold off on merging this until I hear back from them as well. That said, thank you so much for testing, and awesome to hear it works on your end!

If for some reason I don't hear back tomorrow or they reach out saying they won't be able to test, I'll merge. Otherwise as soon as I hear back from them with success, I'll merge.

nomakewan:
Just got word back that they tested it and it worked!

PR #334, opened 2024-07-08
Templates

DustinGrau:
Using GitHub templates to help add some structure to reported issues, starting with Bug Report and Feature Request templates.

### DIFF
--- a/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
+++ b/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
@@ -1,0 +1,77 @@
+name: Bug Report
+description: File a bug report.
+title: "[Bug]: <short description>"
+labels: ["bug", "triage"]
+projects: ["gpstar81/GPStar-proton-pack"]
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thank you for helping to improve this software with a formal issue report.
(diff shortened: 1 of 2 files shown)


DustinGrau:
If you navigate directly to the files, Github will render them in the browser as they are intended though you can't interact with them. It will show fields as required, and as their expected datatype (text, checkbox, etc). If you feel there should be any additions to capture from users, then please let me know.

https://github.com/gpstar81/GPStar-proton-pack/blob/templates/.github/ISSUE_TEMPLATE/BUG-REPORT.yml
https://github.com/gpstar81/GPStar-proton-pack/blob/templates/.github/ISSUE_TEMPLATE/FEATURE-REQUEST.yaml

nomakewan:
Defaulting all issues to asking for a user's e-mail address is not usually something I see in templates.

Perhaps change to `Github, Facebook, E-mail, etc`?

DustinGrau:
I can't recall if you can make an anonymous request (meaning without a Github account) so I thought to add the contact info which was in the sample templates. Perhaps it's not needed as most requests seem to be tied to a real account.

nomakewan:
If an issue is being reported on Github, the user would have to have a Github account, yeah.

nomakewan (approved):
Looks good!

PR #335, opened 2024-07-08
Templates (Redux)

DustinGrau:
Adding 1 more template for general support requests, and disabling the ability to open blank issues. Users will be forced to take a path which uses one of the provided templates, allowing us to require answers to standard questions :)


DustinGrau:
This is where I start to fail with Git, as this branch already had some changes on it so it's picked up the last changes plus the new ones. Still, it should merge just fine since it's non-code items and there's no conflicts.

nomakewan:
Shouldn't the filename for the general support request end in .yaml?

PR #353, opened 2024-07-30
V5.3.2

DustinGrau:
List of changes/fixes:

- Offers a standalone binary for the Attenuator using an Arduino Nano
- Fixes a glitch in the saving of EEPROM data when the inner cyclotron panel is enabled/disabled
- Fixes the vibration motor logic for the Attenuator on ESP32
- Support changing of the default SSID for the ESP32 boards
- Many, many documentation updates with links to revised videos
- Various bugfixes (LED colors, overheating, music playback)
- Support for new 36 pixel inner cyclotron ring
- Cleanup of APIs no longer used/needed
- Add new STL files for integration in non-Haslab packs/wands


DustinGrau:
I need to load the latest binaries on my Haslab and Spirit packs to test things out, though we should be good to go with this bugfix release.

gpstar81 (approved):

nomakewan (approved):
If everything tests good, I don't mind signing off.

There's one thing I'd like to do still (actually been wanting since 5.3.1) but since you already know there'll probably be another v5 release to support the wrist blaster, I can hold off until then to roll it in.

DustinGrau:
Yup, just loaded up your new sound files to test. I'm going to cut this release loose so the cyc panel EEPROM bug is at least fixed, and will likely make 1 more fix branch for 5.3.3 so we have it available.

PR #392, opened 2024-10-08
updated audio firmware

gpstar81:


gpstar81:
Fixes a bug on older versions of GPStar Audio, that plays the test sound when it receives power.

PR #396, opened 2024-10-09
V5.4.1

DustinGrau:
Fixes for ESP32 and making new firmwares for pack, audio official.


nomakewan (approved):
Tested the ESP32 compiles and they worked great. In fact, the momentary glitch when swapping OTAs didn't occur this time around, and the interface remained very snappy.

Also pushed a fix to the 3-LED and 5-LED outer cyclotron users when doing the abbreviated startup from the wand. For this we just skip the ramp and go straight to full speed, like recovering from overheat or button mash.

Doesn't seem like there are any other bugs being reported so going to stamp approval here in case I'm not around in the morning.

DustinGrau:
Latest firmware loaded on my Haslab pack and Spirit pack, the latter being 20-LED cyclotron which I forgot about. All urgent issues appear to be resolved.

Issue #399, opened 2024-10-15
[Bug]: System Default Year Mode via Wand Menu skips current switch year


nomakewan:
### What actions were attempted?

Cycle through all default year mode options via the Wand EEPROM Configuration Menu

### What actually happened?

Default Year Mode -> 1989 -> Afterlife -> Frozen Empire -> Default Year Mode...

### What was expected?

Default Year Mode -> 1984 -> 1989 -> Afterlife -> Frozen Empire -> Default Year Mode...

### Firmware Version

5.4.1

### Mode in Use

Any / Both Modes

### Notable Hardware

N/A

### Homework Completed

- [X] I did my part!

PR #402, opened 2024-10-25
V5.4.2-RC2

DustinGrau:
Resolved Issues:
#399

Ghosts Busted:

-Fixed Afterlife/Frozen Empire wand ramp sound loops and transitions
-Fixed beep when switching wand from normal to standalone mode
-Fixed erroneous pack start if pack is started from wand then ion arm switch is turned on
-Fixed Frozen Empire pack startup not being affected by volume adjustment
-Fixed missing year selection for Default Year Mode in EEPROM Configuration Menu
-Fixed outer cyclotron rapid restart animation on 12-LED or 20-LED setups
-Fixed several sound effect volume levels
-Fixed URL for 50mm fan for DIY smoke kits
-Fixed wifi connectivity issues on mobile devices which expect internet access
-Sanitize user-provided SSID for RFC compliance and mDNS usage
-Several "under the hood" improvements

New Features:

-Added unique Frozen Empire shutdown
-Afterlife/Frozen Empire barrel wing hat light now matches prop behavior
-Updated wand mash error sound effect


DustinGrau:
Opened and ready for feedback

nomakewan:
I've sent the updated ESP32 firmware to one of our customers who had a known issue with Android connectivity and should hear back later today whether it has solved the problem.

On my local testing, this change allowed not only a stable wifi connection that connected quickly, but also allowed my iPhone to continue browsing the web and checking e-mail normally while remaining connected to the ESP32.

nomakewan (approved):
Just heard back from our Android test user, who put this through its paces on multiple Android devices and confirmed that the wifi connectivity issues are solved!! FINALLY!!

I'm going to preemptively approve this. We know that additionally there's a problem where spaces are being allowed in SSIDs which is causing a problem with the ".local" website addressing, but that's a minor change. Everything else looks great!

DustinGrau:
I am working to add code which can sanitize the SSID on the backend, making sure it adheres to the RFC rules for alphanumeric with dashes and underscores, but also remove spaces. The JavaScript will also be adjusted to remove spaces live when the user submits a new name.

DustinGrau:
Code is in, compiled, but not yet tested. Late here for me and I had a long evening, so I'll test it out in the morning unless anyone else can get to it first. The sanitization should occur on user input, but also when pulling a value from the NVS preferences. This may result in an unexpected change in SSID for the user, but this was a gap we should not have allowed.

nomakewan:
Found out why this issue with spaces wasn't caught earlier. The regex checking against the SSID was both the wrong regex and the wrong conditional.

The regex was checking to see if the SSID was a _single character_ that was a-z, A-Z, or an underscore. Then if the SSID _was_ a single character that was a-z, A-Z, or an underscore, it would throw the error saying it was illegal.

I have fixed the regex so that it checks the entire SSID string, and fixed the conditional so that if the test _fails_ it will throw the error saying it was illegal.

DustinGrau:
Thanks for testing this and making the changes to the RegEx. There were obviously some things I though would work, while others did not or weren't needed. I still found some room for improving restrictions on fields that we didn't have, and making sure the formatting is consistent both for page text and how we display error messages.

PR #405, opened 2024-10-27
Fixes for standalone Attenuator, web UI

nomakewan:
This should fix standalone Attenuators so that if for some reason someone flashes the standalone firmware unintentionally, the Attenuator will detect the serial data from the pack and configure itself for full data operation on the fly.

EDIT: Also added a fix for the `printf` partition call to clear up the compiler warnings about the datatype mismatch.


Toy203:
We'll call it the Toy security patch

DustinGrau:
I wonder if we could just make this work the other way (as well). Maybe we start off always assuming the device is usable physically, but until we get a handshake we presume it's disconnected. Once we get any sort of answer to a direct request, or we receive a valid API call it presumes there's a viable connection. This could get rid of the standalone version entirely.

nomakewan:
> I wonder if we could just make this work the other way (as well). Maybe we start off always assuming the device is usable physically, but until we get a handshake we presume it's disconnected. Once we get any sort of answer to a direct request, or we receive a valid API call it presumes there's a viable connection. This could get rid of the standalone version entirely.

It should be, albeit potentially with a minor glitch on initial connection. Basically you would be replacing the "normal" firmware with the Standalone one created using the code from this PR.

Give it a compile, toss it onto your gear, see if it works and how it responds. If it works, then yeah, there can just be the one firmware.

Considering I already did this for the Neutrona Wand (we technically don't need the Bench Test firmware anymore as of 5.3.3), I'm on board. ;)

nomakewan:
Okay, managed to get this working for real this time. Just optimized the `svg` for the pack status screen as far as I could get it, saving 6354 bytes. Thanks to this savings, was able to fit both an `ico` and `svg` favicon for the web UI at a cost of 1488 bytes. So even with the new feature, we still save 4866 bytes. :)

DustinGrau:
> EDIT: Also fixed a bug I noticed while working on this where the ESP wasn't rebooting after an OTA update succeeded. All good now.

Oh shit, that's why the version wasn't getting reflected after update! I noticed that a reboot was needed, but didn't realize why.

DustinGrau (approved):
Everything looks great, and I should be able to run through some tests. I need to merge first so it's in a branch I can access easily, and I may finally have some time to do that today.

PR #407, opened 2024-10-29
Implement audio queueing, bugfixes

nomakewan:
The big change here is the addition of the GPStar Audio firmware 1.0.4 and its "WAV queueing" feature. At present this feature is used to make the Afterlife/Frozen Empire wand ramps more seamless as well as allow the hissing steam effect on the brass pack shutdown to stand alone rather than be cut off by a restart event.

It also introduces a flag called `b_preload_tracks` which, if set to true, will delay playback of all sound files by 50ms in order to give the GPStar Audio more time to buffer the track. While not exposed to the menu system yet, this may resolve issues for users with cheap or damaged SD cards (testing pending arrival of known bad SD cards).

While these new features rely on GPStar Audio 1.0.4, the timers for the original functionality all still exist and will be used as a fallback in case the user is using a WAV Trigger or has not updated their GPStar Audio. Further, if a user for some reason updates the GPStar Audio in their wand but not their pack, this mismatch is properly handled by both the wand and pack firmware.

In addition this brings several bugfixes for 5.4.2 and earlier versions:

-Fixes 35/36 LED inner cyclotron ring animation speed in AL/FE
-Fixes AL/FE wand rampdown sound not playing on Proton Pack
-Fixes Booster Tube smoke machine not activating if N-Filter smoke machine is disabled
-Fixes improper sound sequencing of error states (for example, pulling ribbon cable while overheating)
-Fixes inner cyclotron cake LED not resetting on pack shutdown
-Fixes issue where wand mash sequence was causing incorrect shutdown animation in AL/FE
-Fixes missing wand short startup call in packside wand sound reset function
-Fixes outer cyclotron LED calls remaining active even when lid is removed in 84/89
-Fixes overlapping wand shutdown sound effects during error states
-Fixes pack POST sequence having improper CHSV for stock Haslab LEDs
-Fixes power level adjustment working even if wand is in error state
-Fixes quick vent causing smoke events even if smoke is disabled for N-Filter or Booster Tube
-Fixes smoke machines being stuck on if main smoke switch turned off during smoke event
-Fixes sound transition for button mash lockout sequence
-Fixes sound transition for firing tails
-Adds vent light fadeout to button mash lockout sequence
-Makes it so disabling continuous fire smoke also disables the N-Filter LEDs
-Makes it so power warning LED won't come on if music is playing
-Reduces serial calls for AL/FE wand ramp events

Finally, it also introduces a new "ramp down from full power" sound for AL/FE, and updates our compiler checker to hard define FastLED version 3.7.8 since 3.9.0 is DOA.

Please give this a test and see how it goes. Especially test things like the ribbon cable alarm (for example, yanking it mid-firing, yanking it while overheating, yanking it while the button mash lockout is active). Additionally, please test a pack with either a WAV Trigger or GPStar Audio 1.0.3 to make sure the Afterlife wand ramp sounds don't have any awkward gaps. It should all be working now, but better to test now than be sorry later.


nomakewan:
Okay, finally done taking care of the last things on my list. This isn't 5.4.3 yet (still waiting to test a few things, not to mention we have some Attenuator work left to do) but this should cover all the bugs that have been reported thus far plus the ones I noticed in my own testing.

DustinGrau (approved):
Looks good enough to begin merging and testing. I would suggest an edit to the documentation to clarify when the V104 firmware would be necessary for the GPStar Audio devices, unless we say that the 6.x releases will strongly encourage use of that version while leaving 5.x as-is to either use it or not. Thoughts?

nomakewan:
I agree actually--at the moment the benefit of the updated firmware is just the Afterlife ramp and the brass pack shutdown. There are no other files that benefit from the updated functionality (yet) and so this is not a required update.

But since V6 requires a manual update process for the Attenuator anyway, may as well throw it in.

Issue #408, opened 2024-10-31
[Bug]: Inner Cyclotron way too fast in AL/FE


nomakewan:
### What actions were attempted?

Turn on the pack with the lid off in Afterlife or Frozen Empire mode.

### What actually happened?

Inner cyclotron rotation speed is insanely fast.

### What was expected?

Inner cyclotron rotation speed should be slow enough at idle to perceive a single dot moving in a circle.

### Firmware Version

5.4.2/6.0.Preview

### Mode in Use

Super Hero / Video Game

### Notable Hardware

35 LED Inner Cyclotron Ring.

### Homework Completed

- [X] I did my part!

Toy203:
Honestly thought that was the wanted behavior on this one. It looks cool on mine lol.

nomakewan:
It wasn't intended as far as I'm aware. If no one reported the change I wonder when it actual happened...

Toy203:
Here's mine on 6.0
https://youtu.be/7b0hGVFbi6I

I guess the plexiglass lid could help a bit with the effect, but honestly it looks fine to me. Ramp up and down works well and I like the overall effect.

nomakewan:
> Here's mine on 6.0 https://youtu.be/7b0hGVFbi6I

Which LED ring is this?

Toy203:
The OG 35 ring, afterlife settings.

nomakewan:
Interesting. I too have the OG 35 ring, and while it works fine in 84/89, AL/FE are insanely fast to the point where the LED pattern is broken up pretty badly. I'll have to do some experiments, like introducing the Frutto inner panel into the circuit to see if that has any effect. I've already confirmed that changing the _outer_ cyclotron configuration has zero effect on the behavior of the inner cyclotron, so there's that at least.

Toy203:
Don't get me wrong, I can see the pattern is broken on full rotation speed, leds don't sequentially turn on from what I can see. I just thought it was some kind of trail effect in place or something.

nomakewan:
Nah, it's the actual rotation effect outpacing the FastLED refresh rate. This allows the LEDs to go "faster" but since they're not being refreshed that quickly, LEDs get skipped over.

Issue #410, opened 2024-11-24
[Feature]: Make RGB/GRB toggle for cyclotron cavity available in web UI


nomakewan:
### What would you like to add in terms of software changes?

The Proton Pack code has a configuration setting to allow use of RGB fairy lights as the cyclotron cavity LEDs, but this is currently a compile-time configuration only. This should be added as a simple toggle on the web UI in case a user got RGB LEDs instead of the more common GRB type.

### Would this request involve any specific hardware?

Attenuator/wireless module, inner cyclotron cavity fairy light string

### Homework Completed

- [X] I did my part!

PR #420, opened 2025-01-21
V5.4.3

DustinGrau:
Resolved Issues:
#408
#410
 
Ghosts Busted:
-Fixed 35/36 LED inner cyclotron ring animation speed in AL/FE year modes
-Fixed AL/FE wand rampdown sound not playing on Proton Pack
-Fixed Booster Tube smoke not activating if N-Filter smoke is disabled
-Fixed button mash sequence causing incorrect shutdown animation in AL/FE
-Fixed default volume not going to 100% when using GPStar Audio and no user-set default volume is configured
-Fixed improper client and reboot checks on web interface
-Fixed improper sound sequencing for error states (e.g. yank ribbon cable during overheat)
-Fixed inner cyclotron cake LED not resetting start state on pack shutdown
-Fixed missing wand short startup call in packside wand sound reset sequence
-Fixed outer cyclotron LED calls remaining active even when lid is removed in 84/89
-Fixed overlapping wand shutdown sound effects during error states
-Fixed pack POST sequence using incorrect color value for stock Hasbro LEDs
-Fixed potential vent LED state issues on Single shot Blaster
-Fixed quick vent causing smoke events even if smoke is disabled for N-Filter or Booster Tube
-Fixed smoke machines being stuck on if main smoke switch is turned off during smoke event
-Fixed sound transition for button mash lockout sequence
-Fixed sound transition for the tail end of the firing sequence
-Fixed wand power level adjustment knob working even if in wand error state
-Various "under the hood" fixes and optimizations
 
New Features:
-Added favicon to web interface
-Added new firing sounds for holiday modes
-Added new "ramp down from full power" sound for AL/FE wand
-Added new "sound queueing" feature that makes AL/FE wand sounds smoother
-Added support for GBR, GRB, and RGB cyclotron cavity LEDs
-Added support for GPStar 2-LED mini barrel
-Added support for GPStar 50-LED barrel
-Added support for GPStar addressable RGB vent/top board
-Added support for GRB and RGB inner cyclotron LED rings
-Added vent light fadeout to button mash sequence
-Made it so disabling continuous fire smoke also disables N-Filter LEDs
-Made it so the power idle warning LED won't come on when music is playing
-Separated Halloween and Christmas holiday modes into individual firing modes
-Updated sound effect for button mash lockout to be more accurate to Afterlife
-Updated documentation
 
Note that to get all the benefits of this update, all devices including the GPStar Audio boards will need to be updated.
5.4.3 will still work perfectly well without updating your GPStar Audio boards (or if you are using WAV Triggers), but the new Afterlife/Frozen Empire wand ramp transitions only take effect with updated GPStar Audio firmware.


DustinGrau:
Currently merging code up to the 6.0 preview which must remain in sync as it is still just a pre-release, though I want to test with both if possible to confirm the latest changes work with my older equipment that doesn't yet support the RGB vent.

nomakewan:
Okay, just finished checking over the documentation and found a few discrepancies including with the Github docs, which I've now updated. Here are items that should be updated in the PDF Operational Guide:

Neutrona Wand Menu System Level 2 needs to be updated to match OPERATION.md

EEPROM Configuration Menu Level 1 Option 4 Intensify superscript is "3" but should be "1"

EEPROM Configuration Menu Level 1 Option 2 BWB ring sim description needs to be updated to match OPERATION_EEPROM.md

EEPROM Configuration Menu Level 1 Option 1 Intensify/BWB functions are reversed

EEPROM Configuration Menu Level 2 Option 4 needs to be updated to match OPERATION_EEPROM.md

EEPROM Configuration Menu Level 2 Option 1 Intensify needs to be updated to match OPERATION_EEPROM.md

EEPROM Configuration Menu Level 3 Option 5 BWB is missing Frozen Empire

EEPROM Configuration Menu Level 3 Option 4 BWB needs to be updated to match OPERATION_EEPROM.md

EEPROM Configuration Menu Level 3 Option 2 Intensify needs to be updated to match OPERATION_EEPROM.md

EEPROM Configuration Menu Level 3 Option 1 BWB needs to be updated to match OPERATION_EEPROM.md

DustinGrau:
I noticed on the 6.0 branch there was a missed variable for the JSON data but it looked to have been fixed on the 5.4 branch, so either I missed that or it got fixed already. Either way, the RGB Vent option seems to have no effect on the stock hardware or the Frutto vent light, so that's good. Overall seems to be working on my full pack.

nomakewan:
FastLED 3.9.12 just dropped. It's on the Arduino Library Manager already but hasn't hit the PlatformIO registry yet. Once it does I'll do another binary compile for the checkout. With that plus the fixes to the PDF mentioned earlier I think we'll be ready to go!

DustinGrau:
I made sure my local libraries were up to date and did a recompile. No other changes, but binaries came out different so I checked them in. I've noticed that on the older ESP libraries we still get some warnings from FastLED but that gets resolved on 6.0 when we move to the later esp-idf version for Arduino. We REALLY need to begin moving over to the 6.0 for new fixes/features soon so we can stop building against 2 different sets of libraries. Just something to consider pushing harder in 2025.

DustinGrau:
I'll need to retest since there were modifications due to that volume bug, and I need to try out both the 5 and 6 releases since I changed the esp library in the latter.

DustinGrau:
I'd say the changes are good enough to go out. I ran through a full setup on my pack after clearing the EEPROM and overall it was good. I did notice that while in the LED menu options some voices came from only the wand, only the pack, and some from both--inconsistent but still usable. Also, there seemed to be a lack of voice prompt when cycling to what would be the option for "Inner Cyclotron Panel Dynamic Colors". Though if we're still slating a reworking of the menu system in V6 then there's no worry about fixing that now, and was about the only thing I noticed while in there.

gpstar81 (approved):

PR #431, opened 2025-04-30
Web UI Fix

DustinGrau:
Merges changes for the saturation/hue controls in the web UI for custom colors.


nomakewan (approved):
Looks good! Great work with enforcing saner minimums too.

PR #432, opened 2025-05-01
ESP32 Crash Fix

DustinGrau:
Adjusts the stack size due to a problem recently discovered with the newest espressif (pioarduino) and FastLED libraries which could cause the ESP32 to become unresponsive when the pack was powered up. The exact hardware combination varied as to where the issue would occur but was identified in 2 separate instances among the dev team, which supported claims from the community which were difficult to pin down beyond assumptions of a mis-matched firmware version install.


nomakewan (approved):
LGTM!

PR #436, opened 2025-05-08
Reduce flash tool size

nomakewan:
Reduces all flash tools in size by 50%.


DustinGrau:
Obviously Windows only as you said it was a .Net thing? I'm guessing there was an option for using some kind of shared library rather than bundling direct? Does the reduction in size mean there are any prerequisites on the version of .Net a user must have installed?

DustinGrau (approved):

PR #437, opened 2025-05-08
EEPROM Reset Firmware

DustinGrau:
Implement a new, special firmware which can reset the EEPROM values for an ATMega2560 microcontroller. Intended for the Proton Pack, this could also possibly work for the Neutrona Wand or SingleShot Blaster devices. This is intended for the rare case where the EEPROM may have become corrupted and prevents the device from working normally. It should check whether the EEPROM has already been cleared or not, and perform a full write across the space with zeroes. Since we have limited writes to this space it should only perform this step after checking that the space contains data.


nomakewan (approved):
Looks solid! Great way to wipe the EEPROM as an emergency diagnostic procedure.

PR #438, opened 2025-05-13
Update flash tools

nomakewan:
This updates all flash tools to make them more resilient against user error and edge cases.

GPStar Firmware Flasher:
- Child process will be killed when parent is killed
- Port select will automatically detect and select an FTDI USB device
- avrdude has been updated to fix a potential application hang

GPStar ESP32 Firmware Flasher:
- Child process will be killed when parent is killed
- Invalid COM ports will be automatically filtered out

GPStar Audio Firmware Flasher:
- Child process will be killed when parent is killed
- Port select will automatically detect and select an FTDI USB device


nomakewan:
Note that since the changes are fairly extreme, these flash tools _will_ need to be tested on Windows 10 (and preferably Windows 11 as well) to make sure that the new functions work without throwing any weird exceptions.

Best way is to, for example, plug in your FTDI cable but don't actually connect to a board, and then run the flash tool. Follow the steps to update the firmware as if it were plugged in, click Upload, but then kill the GPStar firmware flasher process using Task Manager. What should happen is not only should the GPStar process end, but avrdude/esptool/stm32flash should also end and no unhandled exceptions should occur in the process.

DustinGrau:
Does this use an updated version of the AVRdude program or are they still debating on the fix and release for that?

nomakewan:
This does indeed use an updated version of avrdude compiled by me that contains all the latest updates and fixes.

gpstar81 (approved):

PR #443, opened 2025-05-30
Correct EEPROM LED checks

nomakewan:
Corrects logic error where EEPROM LED checks were checking against a raw enum rather than an absolute value, causing the checks for the wand barrel LED count and bargraph segment count to fail to read correctly after a power cycle.

This bug was introduced when hardening of EEPROM values was introduced alongside 6.0.1/5.4.Bugfix and as such after being merged with `main` must be upstreamed to `5.4.Fixes` and `6.1.Dev`, then all three branches should have binaries rebuilt and a new 6.0.1 released to replace the existing version.


DustinGrau:
I wonder if it might make sense to force the ENUMs to just be uint8 values? We've used that pattern elsewhere to make sure that each enum has a distinct numeric value rather than relying on the internal numbering. So LEDS_2 = 2, LEDS_5 = 5, etc. Thoughts?

nomakewan:
Really good idea. I had pondered doing that but wasn't sure what potential knock-on effects it would have. After auditing the code it appears there aren't any notable knock-on effects (even the serial data prefs transmission works as-is) so switching this enum paradigm is in fact the better solution.

6.1.x should consider doing this for most (if not all) enums, but that can happen later.

PR #448, opened 2025-06-25
Documentation Cleanup

DustinGrau:
Adjusting the new documentation formatting to use clear heading levels on each page, allowing the navigation sidebar to group items correctly. Adds new text to headers to more easily determine which guide they came from (if possible) when viewing search results. CSS has been adjusted to better fit the table contents for the menus.

Speaking of, this also splits the operation manual into new groups for general usage, runtime menus, and EEPROM menus. This leverages the side navigation to allow users to dive directly into the desired section.


nomakewan (dismissed):
Had a look and the changes look solid and logical. I pushed an update just to fix the header images since the `images` folder was moved into `docs`, so now the header images show up correctly in the markdown files.

Solid update, thank you!

nomakewan:
@DustinGrau In that case, do you have a way so that both the generated pages as well as the original markdown will be viewable without broken images?

I'm going to rescind my approval for now, thank you for bringing that discrepancy to my attention.

nomakewan:
That's totally fine by me; I agree that the formal documentation should be the new implementation. It's just important that we not obviously break the markdown should someone decide to inspect/modify them going forward as they serve as the basis for the new doc deployments.

Had a peek at the changes and this fix works for me!

nomakewan (approved):

DustinGrau:
I'll be keeping this branch open for other changes as needed. I think next I want to address some of the fonts in use, just so I can make the headers more consistent. They're rendering with what looks like a serif font, but should be san-serif to match the navigation. Also, if there's a specific font that Michael uses for his logo I'd like to try to use something similar for better branding.

Issue #426, opened 2025-03-11
[Support]: Music File Troubles.


The-Y33TER:
### What do you need assistance with?

I'm having issues with playing music. I've got all my music files as 500 onward, .wav files & no metadata. Still no dice. What's going on?

### Firmware Version

5.4.3

### Homework Completed

- [x] I did my part!

nomakewan:
We need more information about your music files first.

Please use a tool such as [MediaInfo](https://mediaarea.net/en/MediaInfo/Download) to output text-based specifications for your music files. It should look like this in MediaFire's text output mode:

```
General
Complete name                            : S:\gbsounds\500_library_mono.wav
Format                                   : Wave
Format settings                          : PcmWaveformat
File size                                : 34.4 MiB
Duration                                 : 3 min 24 s
Overall bit rate mode                    : Constant
Overall bit rate                         : 1 411 kb/s

Audio
Format                                   : PCM
Format settings                          : Little / Signed
Codec ID                                 : 1
Duration                                 : 3 min 24 s
Bit rate mode                            : Constant
Bit rate                                 : 1 411.2 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 44.1 kHz
Bit depth                                : 16 bits
Stream size                              : 34.4 MiB (100%)
```

Once you provide some examples of your music files we can proceed.

The-Y33TER:
### Example of the piano jingle I use on every pack I build to indicate music mode has started. 
```
General
Complete name                            : C:\Users\w\Downloads\PackMusic\500_Jingle.wav
Format                                   : Wave
Format settings                          : PcmWaveformat
File size                                : 1.31 MiB
Duration                                 : 7 s 175 ms
Overall bit rate mode                    : Constant
Overall bit rate                         : 1 536 kb/s

Audio
Format                                   : PCM
Format settings                          : Little / Signed
Codec ID                                 : 1
Duration                                 : 7 s 175 ms
Bit rate mode                            : Constant
Bit rate                                 : 1 536 kb/s
Channel(s)                               : 2 channels
Sampling rate                            : 48.0 kHz
Bit depth                                : 16 bits
Stream size                              : 1.31 MiB (100%)

```
### Example 2: The Elmer Bernstein Theme, the first music track I want to play (after the jingle)
```
General
Complete name                            : C:\Users\w\Downloads\PackMusic\501_Bernstein.wav
Format                                   : Wave
Format settings                          : PcmWaveformat
[...]

Issue #451, opened 2025-07-02
[Bug]: EEPROM won't save # of Barrel LEDs


IGendel369:
### What actions were attempted?

Changed the stock Barrel LEDs to the GPstar 50 LED in the EEPROM both using wand and using Wi-Fi web interface after updating to the 6.0.1 software

### What actually happened?

After shutting the battery off and back on again it doesn't remember the settings change. After going back to 6.0 software it saves the setting fine. If I upgrade to 6.0.1 it forgets the setting and won't save it again.

### What was expected?

The EEPROM setting for # of barrel LEDs should be remembered when changing it.

### Firmware Version

6.0.1

### Mode in Use

Super Hero / Video Game

### Notable Hardware

Only upgrade I did was change from Stock Barrel LEDs to the GPstar 50 Barrel LEDs

### Homework Completed

- [x] I did my part!

nomakewan:
Known issue, already fixed in #443 

Feel free to update to the current development binaries in v6.1.Dev if you'd like to test-drive the changes.

Issue #449, opened 2025-06-25
[Feature]: Neutrona Wand Stand Alone


gpstar81:
### What would you like to add in terms of software changes?

-Lower the intensity of the Vibration Motor. 
-Soften / Lower the maximum volume output slightly.

Both of these help with users running AA batteries.

### Would this request involve any specific hardware?

-For stand alone Neutrona Wand users.

### Homework Completed

- [x] I did my part!

nomakewan:
Should be fixed now in v6.1.Dev.

Issue #442, opened 2025-05-28
[Bug]: Adjusting power level quickly after firing causes stuck bargraph pixel.


The-Y33TER:
### What actions were attempted?

I adjusted down my power level from 5 to 1, quickly after I stopped firing.

### What actually happened?

A single pixel in the upper half stayed on when the rest of the bar graph was at level 1/5

### What was expected?

Bargraph runs smoothly with no stuck pixels

### Firmware Version

5.4.4

### Mode in Use

Super Hero / Video Game

### Notable Hardware

28Seg, GPStar Wand Board (all upgrades)

### Homework Completed

- [x] I did my part!

nomakewan:
Was not able to confirm this with the instructions provided, but was able to confirm a similar enough bug that it's likely the same.

While I could not get any stuck bargraph segments when going quickly from 5 to 1, I could indeed get a stuck bargraph segment if I fired in power level 1, then while the animation ramped down switched to power level 2, then powered down to power level 1. There was a very obvious stuck segment. Note that this only appears to happen in Afterlife or Frozen Empire; 84 and 89, in Super Hero mode, do not appear to be affected.

I also noticed that in just the right (wrong?) circumstance, rapidly changing power levels in Afterlife or Frozen Empire could lead to a doubling-up of beep sounds. Something else I will need to look into.

Thank you for the report!

nomakewan:
Found the problem; was a random incorrect integer value that had been introduced alongside the 30-segment bargraph support in v5.4.0, meaning this bug has been around since October of last year and somehow nobody noticed. All fixed now, and also fixed some segment spacing on the 30-segment bargraph that was off.

Thank you again for the report, this is now fixed in the 6.1.Dev branch!

PR #455, opened 2025-07-30
V6.0.2

DustinGrau:
Updates to the latest 3rd-party libraries for builds, bringing devices up to date with the latest software features. Addresses all minor issues and fixes discovered since the v5.4.4 and v6.0 releases. Also includes some housekeeping to rename items for more clarity and consistency with documentation, while fixing said content to remain in agreement and avoid confusion.

Resolved Issues:
#426
#442
#449
#451

Ghosts Busted:
-Fixed bluetooth radio accidentally being enabled on Attenuator/Wifi
-Fixed incorrect power level segment spacing on 30-segment bargraph
-Fixed issue where AL/FE beep loop could double when rapidly changing power level
-Fixed issue where clearing the EEPROM settings on the pack or wand would not always clear settings
-Fixed issue where custom bargraph firing animation EEPROM setting would sometimes be ignored
-Fixed issue where music of specific length could cause auto-skip to fail
-Fixed missing board status LED on single-shot blaster during POST
-Fixed occasional stuck segment issue when ramping down 28-segment bargraph
-Fixed web UI occasionally displaying stale data if pack desync occurs
-Added RGB LED white balancing
-Adjusted AL/FE beep loops to better match on-screen sound and timing
-Disabled AL/FE beep loops while firing to better match on-screen sound
-Disabled RGB powercell by default to better match TVG
-Disabled vibration motor on standalone wands to improve power consumption
-Disabled wand boot errors by default to better match films
-Made power level 5 the default startup power level to better match films
-Reduced automatic music track skip delay by 50%
-Reduced max speaker output on standalone wands to improve power consumption
-Various "under the hood" fixes and optimizations

New Features:
-Added "hot shutdown" feature to allow turning off wand without turning off pack
-Added music playback status to web UI
-Made 84/89 cyclotron cycle rate dynamic based on whether LED fading is enabled or disabled
-Made it so wand will automatically power on in demo light mode if Activate switch is on
-Made RGB inner cyclotron panel the default setting
-Updated all firmware flash tools
-Updated documentation
-Updated STLs


DustinGrau:
Making this PR ready for review, though I have yet to put the updated binaries into any of my actual equipment, so testing is DEFINITELY needed.

nomakewan (approved):
Okay, now that I've had a chance to test the code I've fixed the issues I found, namely that the vibration motor intensity reduction was causing the motor to stall briefly on initial start due to insufficient momentum, which was counterintuitively causing higher current draw than if it were just active at a higher duty cycle.

The vibration feature has now been disabled by default on standalone wands (though it can be turned back on in the EEPROM configuration menu if a user desires), and a new "hot shutdown" feature has been added to address community requests to implement a way to shut down the wand without shutting down the pack.

Looks good on my end!

nomakewan:
Testing looks good on my end; that should do it for changes related to ongoing work in 6.1.Dev as well as user bug reports. LGTM!

DustinGrau:
I just did a full update and gave it a go, went through all the streams on every mode. Made sure the button-mash behavior did what it should (notably for GB:FE) and doing so with and without the lid on the cyclotron. Best I can tell all the usual behavior is there and working, so let's hope we fixed more bugs than we possibly will have caused :)

PR #516, opened 2026-04-28
V6.2.2 - Inline Help

DustinGrau:
First contribution for the v6.2.2 release:
- Adding a new inline help system for web UI's
- Export/Import for settings using JSON files
- Actions now build only on pull requests
- Update version for caching library
- Adjust build Python version
- Update audio serial library


nomakewan (approved):
After that quick pass over the help text to make sure it's in line with the actual functionality, this looks fine to me. Not enough for a full release (plenty more needs to be in before 6.2.2 will be ready) but fine for just the inline help portion.

